PR #3, opened 2018-01-05
Reduce warnings

choefele:
*Description of changes:*
Fixed various warnings when building this project with Xcode 9.2. The main problem was the availability of `NSURLNetworkServiceTypeCallSignaling` on iOS 10.0 only whereas this framework is meant to be compatible with iOS 9.0. This problem causes warnings when integrating this library into apps. Thanks


rohandubal:
Hello @choefele, thanks for submitting the PR. I will have a look at it soon and reply back!

Thanks,
Rohan

PR #18, opened 2018-04-25
Renamed private AWSSRWebSocket classes in order to avoid duplicated symbols when the client project relies on the real SocketRocket

stefanceriu:
AppSync's custom socket rocket clashes with the real one when used together, generating duplicated symbols.

*Description of changes:*

Prefixed internal socket rocket classes with 'AWS'


kvasukib:
Thank you @stefanceriu for reporting to us. We are looking into the issue.

PR #20, opened 2018-05-02
fix sqllite lock issue

raphaelmansuy:
*Issue fixed : the sqlLite database could be locked because the busytimout property on sqllite db connection is by default set to 0

When a lock occurs a crash occurs because there is no retry 

*Description of changes:*

Set db.busytimout  to a positive value in order to retry when a lock occurs in  AWSSQLLiteNormalizedCache.swift


rohandubal:
Thanks for the PR! 😄  I will merge it with the next release.

PR #21, opened 2018-05-02
Renamed private AWSSRWebSocket classes in order to avoid duplicated symbols when the client project relies on the real SocketRocket

stefanceriu:
AppSync's custom socket rocket clashes with the real one when used together, generating duplicated symbols.

Description of changes:

Prefixed internal socket rocket classes with 'AWS'

(Sorry, I closed the other one by mistake when moving some branches around)


rohandubal:
Thanks for the PR! I will merge it with the next release 😄

PR #28, opened 2018-05-30
Replaced spawn of multiple queues in subscriptions for a single queue…

MarioBajr:
Currently batch subscriptions spawn multiple background queues, and apart from the running subscription request, all others sit in the wait state, consuming unnecessary resources and affecting requests from other `AWSAppSyncClient` instances. 
Also for every subscription request, `AppSyncMQTTClient` disconnects all previously connected sockets and connects to the new endpoints, consuming unnecessary resources in case of a batch subscription request.

Description of changes:
- Replaced usage of shared `SubscriptionsOrderHelper` to coordinate sequencial subscription requests with a single queue owned by `AWSAppSyncClient`
- Added a small cancellable delay in `MQTTClient` to avoid sequential disconnections and reconnections of the socket while subscribing to multiple topics.


rohandubal:
Thanks for the PR @MarioBajr ! I am looking into this and will reply back.

MarioBajr:
Awesome, thank you too.

AWSAppSyncClient/AWSAppSyncSubscriptionWatcher.swift
```diff
@@ -64,24 +65,32 @@ public final class AWSAppSyncSubscriptionWatcher<Subscription: GraphQLSubscripti
...
     
     func startSubscription()  {
-        do {
-            while (SubscriptionsOrderHelper.sharedInstance.shouldWait(id: self.uniqueIdentifier)) {
-                sleep(4)
-            }
+        self.subscriptionsQueue.async { [weak self] in
+            let semaphore = DispatchSemaphore(value: 0)
+            
+            self?.performSubscriptionRequest(completionHandler: { (success, error) in
```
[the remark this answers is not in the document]
MarioBajr: You're absolutely right, it should bail out on `self = nil` before acquiring the semaphore. Will submit a fix. The behaviour of not performing the subscription if the watcher gets dealloc is correct though. You should retain the watcher if you have interest in the subscription, releasing it to unsubscribe as you can see in the `deinit`.

rohandubal:
Hello @MarioBajr 

Could you please rebase the changes before I could merge them?

Thanks,
Rohan

MarioBajr:
Done 👍

rohandubal:
Great! Will merge this soon 😄

kvasukib:
@MarioBajr Sorry for the delayed response. Can you resolve the conflicts and update the PR?

MarioBajr:
Hi @kvasukib, sorry for the delay, was on holidays :)
While working on this merge, noticed that a new bug will be introduced in this PR because of a new behaviour added in `AWSIoTMQTTClient`. This PR adds a delay on all `MQTTClient` connections in `AppSyncMQTTClient`, then it's possible to have a watcher being deallocated before any attempt to connect. The problem is that `AWSAppSyncSubscriptionWatcher` tries to unsubscribe on `deinit`, triggering the new exception `Cannot call unsubscribe before connecting to the server` inside `unsubscribeTopic:ackCallback:` because `_userDidIssueConnect`is FALSE. Will work to solve this issue soon.

MarioBajr:
Done, had to skip `AWSAppSyncTests.swift` from running in travis because it requires some configuration. About these tests, it behaves as integration tests, having mixed feelings if it should be in this SDK. I would replace them by unit tests with responses coming from a mocked networkTransport.

rohandubal:
@MarioBajr I was trying to integrate these changes, but I am running into issues while testing on my side. Whenever I have more than 1 subscription, it was not working. Have you seen any such behavior on your side? I want to get the improvements in too so want to determine the best path ahead.

AWSAppSyncClient/AppSyncMQTTClient.swift
```diff
@@ -91,59 +99,96 @@ class AppSyncMQTTClient: AWSIoTMQTTClientDelegate {
...
+                        self.dictionary[topic] = watchers
+                    }
+                }
+            }
+        }
+        
+        func remove(subscription: MQTTSubscritionWatcher) {
+            synchronized() {
+                dictionary.forEach({ (element) in
+                    element.value.allObjects.filter({ $0.getIdentifier() != subscription.getIdentifier() }).forEach({ (watcher) in
```
rohandubal: I think this should be `==` instead of `!=`

```swift
element.value.allObjects.filter({ $0.getIdentifier() == subscription.getIdentifier() }).forEach({ (watcher) in
```

.travis.yml
```diff
@@ -13,17 +13,17 @@ before_deploy:
 - carthage build --no-skip-current
 - carthage archive $FRAMEWORK_NAME
 script:
-- xcodebuild -workspace AWSAppSyncClient.xcworkspace -scheme AWSAppSync build test -destination 'platform=iOS Simulator,name=iPhone 6,OS=latest'
+- xcodebuild -workspace AWSAppSyncClient.xcworkspace -scheme AWSAppSync build
```
MarioBajr: Any reason to remove tests from running in travis to validade PRs?

MarioBajr:
Nice catch, will add some tests covering this before submitting the fix.

rohandubal:
Awesome! I also identified an issue with the existing integration(functional) tests:

https://github.com/awslabs/aws-mobile-appsync-sdk-ios/blob/master/AWSAppSyncTests/AWSAppSyncTests.swift#L231

Here since we are doing `let _ = subscribe` and not capturing the result, the object is getting de-allocated (rightfully so). The fix in the tests is to hold a strong reference to the returned object. I think we are nearing the merge 😄

.travis.yml
rohandubal: It is definitely a good idea to run the tests on travis. I am working to add some encrypted variables(API URL info, auth information) which would help run the functional tests as well. I will enable along with them 😃

PR #30, opened 2018-05-30
Calling the AWSAppSyncClientConfiguration connectionStateChangeHandler

stefanceriu:
Calling the AWSAppSyncClientConfiguration connectionStateChangeHandler on network availability changes and using .Online when the endpoint is reachable.


minbi:
Looking into merging this

PR #32, opened 2018-06-05
Preventing IAM signature request from blocking the main thread.

stefanceriu:
The current implementation of AWSAppSyncHTTPNetworkTransport's send methods block the main thread waiting for the AWSSignatureV4Signer's credentials provider to fetch access tokens. This is particularly noticeable under poor network conditions.

This PR wraps the URLSessionDataTasks in a Cancellable AWSAppSyncHTTPNetworkTransportOperation and returns early so it doesn't block anymore. The internal dataTask will be populated later as a result of the AWSTask chain.


rohandubal:
Thanks for the PR @stefanceriu !

I will have a look at this and reply back.

stefanceriu:
Awesome, thank you too. We'll add a small commit to it soon to support cancellations while requesting tokens.

rohandubal:
That would be great! Will wait for it 😄

Issue #38, opened 2018-06-21
Mock Response from server to allow for Tests against mock data


joelmbell:
Wondering if there is a sanctioned way to mock out data responses from the server so we can implement UI Tests against that mock data.

I've tried a couple different approaches so far. The first is setting up a custom URLSessionConfiguration with a Mock URL Cache that just returns mock data whenever the URL cache is checked by the URLSession. I haven't quite gotten this to work yet, and it seems a bit hacky, but is a possible solution.

The other thing I'm thinking about trying is modifying the AWSAppSyncClientConfiguration to expose the ApolloCache so our application can inject a custom ApolloCache implementation that returns mock data. This seems to be the preferred way of mocking data in Apollo (https://github.com/apollographql/apollo-ios/issues/188)

The second approach would require a modification to the AppSync framework. I'd be happy to submit a PR for this, but wanted to make sure there wasn't some other approach I haven't considered yet.

joelmbell:
An even better solution might be to expose the ability to configure the NetworkTransport used by Apollo. Then we can just return mock data straight from there. I'm exploring this option now

PR #40, opened 2018-06-26
AWSNetworkTransport modifications to enable the ability to mock server

joelmbell:
Issue #38 

The purpose of this PR is to expose the NetworkTransport class that is used by the App Sync SDK so it is possible to create a Mock Network Transport object to control what responses from the server look like in our Functional UI tests.

*Description of changes:*

* Creating a protocol outlining the requirements for a NetworkTransport class

* Modifying internal dependecies on the concrete AWSAppSyncHTTPNetworkTransport class, so they are only depedent on the protocol

* Exposing a new configuration initializer for app sync where the app can create their own network transport class. This includes a small refactor to the configuration class.


rohandubal:
Thanks for the PR @joelmbell ! I will review this shortly and get back to you.

PR #53, opened 2018-08-12
Allow Auth credentials to be fetched via callback

JohnRbk:
For OIDC or Cognito authentication, auth tokens often expire and need to be re-generated. Using a callback follows the model also used by aws-mobile-appsync-sdk-js


rohandubal:
Hello @JohnRbk 

Thanks for the PR! This is a good use-case which we can add to the SDK. But the way the current changes are, this would be a breaking change for the consumers. Is there a non-breaking solution which you can think of and update the PR? 

I will discuss with the team about the possibility of introducing a breaking change in the meanwhile.

Thanks,
Rohan

JohnRbk:
Hi Rohan, yes, I think there is a way to make this a non-breaking change. I do think it has trade-offs with respect to the programming model; even on the Android sdk I've seen examples where people have used semaphores to work around the lack of a callback, and that leads to overly complicated code. I think it would also make sense to align a solution across all platform SDKs.

Regardless, here is a way to make this a non-breaking change using a new protocol. Let me know what you think. (Maybe a fatalError() in the extension `getLatestAuthToken` would be better to avoid improper use).

In file AWSAppSyncAuthProvider.swift:

```swift
public protocol AWSOIDCAuthProviderAsync: AWSOIDCAuthProvider {
    func getLatestAuthToken(_ callback: @escaping (String) -> Void)
}
extension AWSOIDCAuthProviderAsync {
    public func getLatestAuthToken() -> String { return "" }
}
```

And in AWSAppSyncHTTPNetworkTransport.swift:

```swift
    private func sendRequestWithAuth(mutableRequest: NSMutableURLRequest, sendRequest: @escaping (URLRequest) -> Void ) {        
        switch self.authType {
            
        case .awsIAM:
            let signer:AWSSignatureV4Signer = AWSSignatureV4Signer(credentialsProvider: self.credentialsProvider, endpoint: self.endpoint)
            signer.interceptRequest(mutableRequest).continueWith { _ in
                return nil
                }.continueWith { _ in
                    sendRequest(mutableRequest as URLRequest)
            }
        case .apiKey:
            mutableRequest.setValue(self.apiKeyAuthProvider!.getAPIKey(), forHTTPHeaderField: "x-api-key")
            sendRequest( mutableRequest as URLRequest)
        case .oidcToken:
            
            if let authProvider = self.userPoolsAuthProvider as? AWSOIDCAuthProviderAsync {
                authProvider.getLatestAuthToken { token in
                    mutableRequest.setValue(token, forHTTPHeaderField: "authorization")
                    sendRequest( mutableRequest as URLRequest)
                }
            }
            else if let authProvider = self.userPoolsAuthProvider {
                mutableRequest.setValue(authProvider.getLatestAuthToken(), forHTTPHeaderField: "authorization")
                sendRequest( mutableRequest as URLRequest)
            }
            
        }
    }
```

joelmbell:
I could really use this, any chance of it getting merged in soon?

kvasukib:
@JohnRbk The PR has conflicts. I have touched the same pieces of code in my PR and that's the reason you may have conflicts. Can you resolve the conflicts and apply your non-breaking changes to the PR?

@joelmbell I'm waiting on @JohnRbk to resolve the conflicts. I will do a final review once the conflicts are resolved.

AWSAppSyncClient/AWSAppSyncAuthProvider.swift
```diff
@@ -8,7 +8,7 @@ import Foundation
 // For using OIDC based authorization, this protocol needs to be implemented and passed to configuration object.
 public protocol AWSOIDCAuthProvider {
     /// The method should fetch the token and return it to the client for using in header request.
-    func getLatestAuthToken() -> String
+    func getLatestAuthToken(_ callback: @escaping (String) -> Void)
```
kvasukib: Can you make this a non-breaking change?

JohnRbk:
Hey guys, I re-baselined my project with the latest master & have applied my updates. Let me close out this PR and issue a new clean one. I also have some unit tests to add as well.

Issue #15, opened 2018-04-06
Carthage build fails


joelmbell:
When building with Carthage, the AWSAppSync build script fails. It looks to be in relation to some Cocoapods dependencies.

I have removed the ```/Users/joelbell/Library/Caches/org.carthage.CarthageKit``` folder and the ```Carthage/``` folder from my project and I still see this issue.

This is my Carthage update command:
```carthage update --platform ios --color auto --no-use-binaries```

Below is a portion of the error log that is relevant.

```
PhaseScriptExecution [CP]\ Check\ Pods\ Manifest.lock /Users/joelbell/Library/Caches/org.carthage.CarthageKit/DerivedData/9.3_9E145/aws-mobile-appsync-sdk-ios/2.6.15/Build/Intermediates.noindex/ArchiveIntermediates/AWSAppSync/IntermediateBuildFilesPath/AWSAppSyncClient.build/Release-iphoneos/AWSAppSync.build/Script-4E996121F39BA74DB553A2CB.sh
    cd /Users/joelbell/dev/SIC/abx-sic-operator/Carthage/Checkouts/aws-mobile-appsync-sdk-ios
    /bin/sh -c /Users/joelbell/Library/Caches/org.carthage.CarthageKit/DerivedData/9.3_9E145/aws-mobile-appsync-sdk-ios/2.6.15/Build/Intermediates.noindex/ArchiveIntermediates/AWSAppSync/IntermediateBuildFilesPath/AWSAppSyncClient.build/Release-iphoneos/AWSAppSync.build/Script-4E996121F39BA74DB553A2CB.sh
diff: /Podfile.lock: No such file or directory
diff: /Manifest.lock: No such file or directory
error: The sandbox is not in sync with the Podfile.lock. Run 'pod install' or update your CocoaPods installation.

** ARCHIVE FAILED **


The following build commands failed:
	PhaseScriptExecution [CP]\ Check\ Pods\ Manifest.lock /Users/joelbell/Library/Caches/org.carthage.CarthageKit/DerivedData/9.3_9E145/aws-mobile-appsync-sdk-ios/2.6.15/Build/Intermediates.noindex/ArchiveIntermediates/AWSAppSync/IntermediateBuildFilesPath/AWSAppSyncClient.build/Release-iphoneos/AWSAppSync.build/Script-4E996121F39BA74DB553A2CB.sh
(1 failure)
```

rohandubal:
[...]

PR #58, opened 2018-08-21
Migrate project dependency manager to Carthage

juankaram:
*Issue #, if available: #15*

*Description of changes:*
I feel this pull request maybe over the top but I would like to submit it for your consideration.

This effectively changes the dependency manager to Carthage, which helps the community to build this library with the latest Swift toolchain (or even beta) thanks to the way Carthage manage its packages, but also consume packaged binaries if available for the current release. 

The downsize is that if you choose to build a different toolchain that the one pre packaged, the initial bootstrap times could be large due the dependency on AWSCore, how aws-ios-sdk is structured and Carthage lack of support to ignore schemas.

If you want to give it a spin just download the PR and `carthage update --platform iOS --no-use-binaries` will build the dependencies for the toolchain configured in Xcode-Select or Carthage environment variable, allowing the project to effectively be compiled with swift 4.2. If you want a faster bootstrap you can always use `carthage update --platform iOS` and it will use the latest packed binaries available.

As a note, had to change the Xcode project due the required dependency configuration changes. I hope this change didn't break the environment of the community cocoapods users, the intention is to support both.

Possibly I will need some help checking if the update on the travis-ci script is working. I was not able to test it on my side because I don’t have the appropriate secrets.

Please feel free to decline merge this pull request, there is no obligation for any of the maintainers to merge this changes.


palpatim:
@juankaram, thanks for the thought you put into this issue and for the PR. Although you make some good points about the benefit of using Carthage, we've decided to reject the PR and retain CocoaPods as the dependency manager for the AppSync project itself (we will of course continue to vend both Podspecs and Carthage binaries so users of either platform can consume the framework).

The primary reason for retaining CocoaPods is that it speeds the development process of the AppSync framework itself. As you point out, migrating to Carthage would potentially gain us the ability to use dependency binaries which are compatibile with the version of Swift we're currently using. However:
1. Developers still have to re-build if the packaged Carthage binaries are not built with the same Swift version (which you noted as a downside above)
1. We lose the ability to easily edit those dependencies concurrently with AppSync framework development, which for the AppSync project is an important part of the development cycle. With Carthage, we would have to make updates in a separate project and manually build; with CocoaPods we can link to the dependency project via a local `:path` and have it automatically re-build with the AppSync project build.

I didn't fully understand one point you raised:

> [This change] helps the community to build this library with the latest Swift toolchain (or even beta) thanks to the way Carthage manage its packages

If my understanding is correct, CocoaPods would also allow developers to build with whatever toolchain they have installed, as the dependency source is packaged directly into the source of the project.

Please let me know if I'm misunderstanding your point.

PR #61, opened 2018-08-24
Added checks to ensure that the SDK doesn't crash if the received mes…

cbommas:
Added checks to ensure that the SDK doesn't crash if the received message is not convertible to String or JSONObject.

Added a logger component to log the error message.

*Issue #, if available:*

*Description of changes:*


rohandubal:
Updated the PR with logic to share logging client with `AWSCore`

Issue #42, opened 2018-07-02
Enhancements: It would be good to expose the detailed connection states to the user


fans3210:
It would be good to expose the detailed connection states to public so that user can choose to manually fetch the missing messages between the disconnect-to-connected time window. Eg: Connecting, Connected, Not Observable(might be 'connecting' too although might not be observable), Disconnected etc.

rohandubal:
Hello @fans3210 

Thanks for the feedback. Can you elaborate a little more? 

For e.g. Do you want these states at a client level or at a subscription level?  Can you elaborate the use-cases which you are targeting?

Thanks
Rohan

fans3210:
@rohandubal . Thanks. I prefer this to be on server side. Eg: 1. If I use lambda resolver, I prefer I can get all the subscribers' connection status in the lambda function's event field like event.subscriberStatus which is like a dictionary or json   2. If I'm using other types of resolvers beside lambda one, I can get the dictionary/json from the resolver's template code like: $context.subscriberStatus. That's only my personal opinion. I don't know where is the best place to store such information. But it should be on server side since I can choose whether or not to send push notification in lambda functions.

MarioBajr:
Hi @fans3210, I have a branch fixing this, waiting #35 to be merged to create the PR

fans3210:
@MarioBajr . Amazing. Tks

muellerfr:
@rohandubal Looks like this PR was merged a while back. Can this issue be closed?

MarioBajr:
Hi @muellerfr, 
I believe that you're referring to this PR #75 , and it hasn't been merged yet. In order to create the observable subscriptions status PR on top of the master branch I need the #28 to be accepted and merged.

palpatim:
@fans3210 

Regarding your use case description:

> UseCase:
> For a chat app, normally I will unsubscribe all topics when the app entered background to reduce the battery and data cost. By exposing the detailed connection status to the developer, when app enters foreground again, I can manually handle retrieval of the chat messages I missed during the disconnection period.

AppSync currently suspends network activity when it enters the background, and resumes when the app resumes foreground. In addition, the recently-added [Delta Sync](https://aws-amplify.github.io/docs/ios/api#delta-sync) feature basically does just what you specify: when creating a new `sync` subscription, the system will query for only the changes since the last successful sync.

Does that meet your needs for this use case?

[...]

PR #75, opened 2018-09-22
Added SubscriptionStatusObserver block to subscribe calls in AWSAppSyncClient

MarioBajr:
The current AppSyncClient subscription API provides only callbacks to receive real time updates, but not enough to provide a better experience. The only callback that can be registered is the `resultHandler`, which provides feedback from failure or the real time message package, not exposing basic connection status like if I managed to subscribe and just sits in the connected state waiting for updates, or if I'm still connecting, not being ready to receive real time updates.

The goal of this PR is to expose the status of subscriptions, enabling clients to know when the system is ready to receive real time updates. Not only exposing all steps while establishing a connection but also all different types of errors.

*Description of changes:*
- Created a new optional block in the `subscribe` method, requesting the closure `SubscriptionStatusObserver` that will provide updates, and the property `status` to the `MQTTSubscriptionWatcher ` protocol.
- Deprecated the unnecessary `getIdentifier()`, relying on the behaviour that `MQTTSubscriptionWatcher` is locked to be reference typed.
- Replaced generic struct based `AWSAppSyncSubscriptionError` by a more descriptive enum based.
- Updates MQTTClient tests

*Considerations:*

This code deprecates:
- `MQTTSubscriptionWatcher.getIdentifier`
- `AWSAppSyncSubscriptionWatcher.uniqueIdentifier`
- `AWSAppSyncSubscriptionError.additionalInfo`
- `AWSAppSyncSubscriptionError.errorDetails`

This code removes:
- `class SubscriptionsOrderHelper`
  
### DIFF
--- a/AWSAppSyncClient/AWSAppSyncClient.swift
+++ b/AWSAppSyncClient/AWSAppSyncClient.swift
@@ -18,6 +18,8 @@ public protocol ConnectionStateChangeHandler {
 
 public typealias SubscriptionResultHandler<Operation: GraphQLSubscription> = (_ result: GraphQLResult<Operation.Data>?, _ transaction: ApolloStore.ReadWriteTransaction?, _ error: Error?) -> Void
 
+public typealias SubscriptionStatusObserver = (_ status: SubscritionWatcherStatus) -> Void
+
 public typealias OptimisticResponseBlock = (ApolloStore.ReadWriteTransaction?) -> Void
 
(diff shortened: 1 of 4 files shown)


rohandubal:
Hello @MarioBajr 

Thanks for this PR and sorry I have not been able to get to it. Exposing these many states might make it a bit complicated for everyone to use. We need to refine how we want to make this functionality available. I am closing this PR in favor of coming up with a spec discussion/ feature request on how to support this in #42 

Thanks,
Rohan

Issue #79, opened 2018-10-06
Memory leak when using AppSyncClient


Hpark11:
**Describe the bug**
Memory leak continuously occurred when using AppSyncClient (especially when invokes mutation)

**To Reproduce**
1. use appsyncClient.perform()

**Expected behavior**

**Screenshots**
<img width="1440" alt="2018-10-06 7 52 24" src="https://user-images.githubusercontent.com/8099899/46570615-cfeaf680-c9a1-11e8-89d2-47fc9d2cc3bb.png">

**Environment(please complete the following information):**
 - AppSync SDK Version: 2.6.22
 - Dependency Manager: Cocoapods
 - Swift Version : 4.1

**Device Information (please complete the following information):**
 - Device: iPhone7, Simulator
 - iOS Version: iOS 11.4
 - Specific to simulators:

**Additional context**

ifabijanovic:
I'm afraid it's worse than this - the client itself is in a retain cycle from initialization and it cannot be released (on logout for example). Simply doing something like this (using code from sample):

```   
let databaseURL = URL(fileURLWithPath:NSTemporaryDirectory()).appendingPathComponent("events-app-db")            
let appSyncConfig = try AWSAppSyncClientConfiguration(appSyncClientInfo: AWSAppSyncClientInfo(), databaseURL: databaseURL)
self.appSyncClient = try AWSAppSyncClient(appSyncConfig: appSyncConfig)
self.appSyncClient = nil
```

Will not release `AWSAppSyncClient`, which in turn won't release a whole graph of objects including an SQLite Connection.

joelmbell:
This is affecting us as well, hoping that MR can get merged into a new version soon. We currently have a bug where mutations are attempting to send after a user has logged out. AppSyncClient should be nil'd out before that happens but due to the retain cycle we can't stop it from sending stored mutations if they exist

palpatim:
Hi folks,

We've recently added tests to ensure that we do in fact release `appSyncClient` when the owner sets to `nil`. 

https://github.com/awslabs/aws-mobile-appsync-sdk-ios/commit/4936c31af45a89eaaade78752499f686c50fc1c6

It should be fixed as of 2.9.1.

PR #88, opened 2018-10-23
Resolved retain cycles in AWSAppSyncClient

ifabijanovic:
Partially related to issue #79.

Problem:
Class `AWSAppSyncClient` sets up 2 retain cycles in its initializer which prevent deallocation.

Solution:
1. Cycle in `MutationExecutor` broken by using a `weak` reference
2. Cycle in `AWSAppSyncClient` itself broken by not adding `self` to `networkStatusWatchers` array but instead just invoking method on self directly later in code.

Verification:
A new unit test confirming `AWSAppSyncClient` gets deallocated.


shannon-hager-skookum:
I can't speak to issues outside of our specific use case but this fix frees up over half of the usable RAM on our devices that is never freed up due to the retain cycles it addresses.

PR #100, opened 2018-11-15
Initial changes for subscriptions updates.

rohandubal:
*Description of changes:*

Improve subscriptions behavior.


AWSAppSyncClient/AWSAppSyncClient.swift
```diff
@@ -954,16 +972,17 @@ public class AWSAppSyncClient {
...
-    ///   - query: The query to fetch.
-    ///   - cachePolicy: A cache policy that specifies when results should be fetched from the server and when data should be loaded from the local cache.
-    ///   - queue: A dispatch queue on which the result handler will be called. Defaults to the main queue.
-    ///   - resultHandler: An optional closure that is called when query results are available or when an error occurs.
-    ///   - result: The result of the fetched query, or `nil` if an error occurred.
-    ///   - error: An error that indicates why the fetch failed, or `nil` if the fetch was succesful.
-    /// - Returns: An object that can be used to cancel an in progress fetch.
+    ///   - baseQuery: The base query to fetch which contains the primary data.
+    ///   - baseQueryResultHandler: Closure that is called when base query results are available or when an error occurs. Every time a sync operation is called, a fetch for the baseQuery from the cache will be done first before initiating any other operations.
+    ///   - subscription: The subscription query which will provide real time updates.
```
palpatim: 1. Is it a subscription "query" or a subscription request with callbacks, etc? It's not clear from the docs what the relationship between `subscription` and `deltaQuery` are.
1. The number of parameters in this signature makes me uncomfortable. We have a number of places where we pass GraphQL operations and callbacks; would it make sense to abstract those into a single type that can encapsulate these closely-coupled behaviors?

AWSAppSyncClient/AWSAppSyncClient.swift
```diff
@@ -631,8 +647,65 @@ public protocol AWSAppSyncOfflineMutationDelegate {
...
+    var allowsCellularAccess: Bool = true
+    var isInitialConnection: Bool = true
+    
+    static func setupSharedInstance(host: String, allowsCellular: Bool) {
+        sharedInstance = AWSAppSyncNetworkStatusChangeNotifier(host: host, allowsCellular: allowsCellular)
+    }
+    
+    static var sharedInstance: AWSAppSyncNetworkStatusChangeNotifier?
+    
+    init(host: String, allowsCellular: Bool) {
```
palpatim: If this is a singleton, should the `init` be private?

AWSAppSyncClient/AWSAppSyncClient.swift
```diff
@@ -631,8 +647,65 @@ public protocol AWSAppSyncOfflineMutationDelegate {
...
+            if (self.allowsCellularAccess) {
+                isReachable = true
+            }
+        case .none:
+            isReachable = false
+        }
+        
+        let info = AppSyncConnectionInfo.init(isConnectionAvailable: isReachable, isInitialConnection: isInitialConnection)
+        
+        if (isInitialConnection) {
```
palpatim: It looks like this should be a `guard !isInitialConnection ...` at the top of the method?

AWSAppSyncClient/AWSAppSyncClient.swift
```diff
@@ -757,6 +829,18 @@ public class AWSAppSyncClient: NetworkConnectionNotification {
                                               resultHandler: resultHandler)
     }
     
+    internal func subscribeWithConnectCallback<Subscription: GraphQLSubscription>(subscription: Subscription, queue: DispatchQueue = DispatchQueue.main, connectCallback: @escaping (() -> Void), resultHandler: @escaping SubscriptionResultHandler<Subscription>) throws -> AWSAppSyncSubscriptionWatcher<Subscription>? {
```
palpatim: Here and throughout: I believe this queue is used to invoke callbacks, right? If so, can we clarify the name to something like `callbackQueue`--right now, the name implies that this is the queue on which the subscription requests and processing will be managed. Either way, is the `main` queue the right default for this? Why wouldn't we create a private queue to be used as a default?

AWSAppSyncClient/AppSyncSubscriptionWithSync.swift
```diff
@@ -0,0 +1,466 @@
...
+            
+        }
+        return success
+    }
+    
+    
+    /// Runs the delta query based on the given criterias.
+    ///
+    /// - Returns: true if the operation is executed successfully.
+    func runDeltaQuery() -> Bool {
```
palpatim: Is the return value necessary? It never returns anything but `true`

AWSAppSyncClient/AWSAppSyncSubscriptionMetadataCache.swift
```diff
@@ -0,0 +1,62 @@
...
+internal final class AWSSubscriptionMetaDataCache {
+    
+    private let db: Connection
+    private let subscriptionMetadataRecords = Table("subscription_metadata")
+    private let id = Expression<Int64>("_id")
+    private let operationHash = Expression<String>("operationHash")
+    private let lastSyncDate = Expression<Date?>("lastSyncDate")
+    
+    public init(fileURL: URL) throws {
+        db = try Connection(.uri(fileURL.absoluteString), readonly: false)
```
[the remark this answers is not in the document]
rohandubal: This is intentional. Customers can use different cache paths for different AppSync clients. There is only 1 instance of SubscriptionMetadataCache which is shared across all SyncWatchers and is owned by AppSync client.

AWSAppSyncClient/AWSAppSyncSubscriptionWatcher.swift
```diff
@@ -21,6 +21,7 @@ import os.log
     func getTopics() -> [String]
     func messageCallbackDelegate(data: Data)
     func disconnectCallbackDelegate(error: Error)
+    func connectedCallbackDelegate()
```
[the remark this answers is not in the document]
rohandubal: Fixed :)

AWSAppSyncClient/AWSAppSyncClient.swift
rohandubal: 1. Fixed documentation to make it clear.
2. It would definitely be a good idea to move the operations and their completion handlers into a single entity, but due to the dependency we have it is difficult and would require a lot of changes. Will discuss this offline.

AWSAppSyncClient/AWSAppSyncClient.swift
rohandubal: Yup, fixing it.

AWSAppSyncClient/AWSAppSyncClient.swift
rohandubal: Fixed.

AWSAppSyncClient/AWSAppSyncClient.swift
rohandubal: Agreed on the naming change. The default is main queue for the assumption that developer would want to do a UI update when the data is available. We can discuss about changing the default queue for this method.

AWSAppSyncClient/AWSAppSyncClient.swift
```diff
@@ -809,6 +893,120 @@ public class AWSAppSyncClient: NetworkConnectionNotification {
...
+    public func sync<BaseQuery: GraphQLQuery>(
+        baseQuery: BaseQuery,
+        baseQueryResultHandler: @escaping OperationResultHandler<BaseQuery>,
+        queue: DispatchQueue = DispatchQueue.main,
+        syncConfiguration: SyncConfiguration = SyncConfiguration.defaultSyncConfiguration()) -> Cancellable {
+        let subs = EmptySubscription.init()
+        let subsCallback: (GraphQLResult<EmptySubscription.Data>?, ApolloStore.ReadTransaction?, Error?) -> Void =  { (res, trans, err) in
+        }
+        let deltaQuery = EmptyQuery.init()
+        let deltaCallback: (GraphQLResult<EmptyQuery.Data>?, ApolloStore.ReadTransaction?, Error?) -> Void =  { (res, trans, err) in
```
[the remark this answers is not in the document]
rohandubal: Good callout; fixed.

AWSAppSyncClient/AWSAppSyncHTTPNetworkTransport.swift
```diff
@@ -183,13 +184,16 @@ public class AWSAppSyncHTTPNetworkTransport: AWSNetworkTransport {
                     case .success(let jsonBody):
                         completionHandler(jsonBody, nil)
                     case .failure(let error):
+                        let taskUUID = UUID().uuidString
                         let (shouldRetry, backoffTime) = retryHandler.shouldRetryRequest(for: error)
                         if (shouldRetry && backoffTime != nil) {
```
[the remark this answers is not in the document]
rohandubal: Fixed.

AWSAppSyncClient/AWSAppSyncRetryHandler.swift
```diff
@@ -17,10 +17,15 @@ internal class AWSAppSyncRetryHandler {
...
     func shouldRetryRequest(for error: AWSAppSyncClientError) -> (Bool, Int?) {
-        let httpResponse: HTTPURLResponse?
+        currentAttemptNumber += 1
+        var waitMillis = Int(Double(truncating: pow(2.0, currentAttemptNumber) as NSNumber) * 100.0 + Double(AWSAppSyncRetryHandler.getRandomBetween0And1() * AWSAppSyncRetryHandler.JITTER))
+        
+        var httpResponse: HTTPURLResponse?
+
         switch error {
-        case .requestFailed(_, let reponse, _):
+        case .requestFailed(_, let reponse,_):
```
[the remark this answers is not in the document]
rohandubal: Fixed

AWSAppSyncClient/AppSyncSubscriptionWithSync.swift
```diff
@@ -0,0 +1,466 @@
+//
+//  AWSAppSyncDeltaSubscription.swift
+//  AWSAppSync
+//
+
+import Foundation
+
+public enum DeltaSyncState {
```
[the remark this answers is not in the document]
rohandubal: I will rename to `SyncState`

AWSAppSyncClient/AppSyncSubscriptionWithSync.swift
rohandubal: It is a miss. should return false on a failed response. Fixing.

AWSAppSyncClient/AppSyncMQTTClient.swift
```diff
@@ -81,6 +85,16 @@ class AppSyncMQTTClient: AWSIoTMQTTClientDelegate {
     }
     
     private func resetAndStartSubscriptions(subscriptionInfo: [AWSSubscriptionInfo], identifier: String){
```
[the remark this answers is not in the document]
rohandubal: Added

AWSAppSyncClient/AppSyncSubscriptionWithSync.swift
```diff
@@ -0,0 +1,466 @@
...
+                  subscriptionMetadataCache: AWSSubscriptionMetaDataCache?,
+                  syncConfiguration: SyncConfiguration,
+                  handlerQueue: DispatchQueue) {
+        self.appsyncClient = appsyncClient
+        self.subscriptionMetadataCache = subscriptionMetadataCache
+        self.syncConfiguration = syncConfiguration
+        self.handlerQueue = handlerQueue
+        self.baseQuery = baseQuery
+        
+        if Subscription.operationString != "No-op" {
```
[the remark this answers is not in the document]
rohandubal: I could not make them as optional since Swift compiler would not be able to figure the type of the generic parameters ([Ref](https://stackoverflow.com/questions/28458160/default-value-for-optional-generic-parameter-in-swift-function))

I will refactor `No-op` logic.

Issue #36, opened 2018-06-14
No way to clear client cache


michaeleconomy:
Ok, so i'm struggling with this case:

i have a query that does "get info about the current user".  It doesn't take an id as a parameter, instead depends on the users context.identity.cognitoIdentityId (iam auth).

The problem i have is: if you do that call, i believe it is caching the response, then if i sign out and sign back in (as a different user), i get the old user's information when i do the same call.

Perhaps i can play with the cache keys (so it contains the user id in it)?  - i'm not sure how to set this:
appSyncClient?.apolloClient?.cacheKeyForObject = { $0["id"] }

I didn't see documentation for this method, but maybe i'm missing it?

Alternatively - it would be great to just expose appSyncClient.store.clearCache (i think this is a thing here: https://github.com/apollographql/apollo-ios/blob/master/Sources/Apollo/ApolloStore.swift)

Sorry if i'm not making sense - I'll be happy to go into more detail if needed.

michaeleconomy:
Alternatively, if there was a way to disable caching entirely, that would work also - I'd want to do caching at a higher level anyway - so request caching doesn't make sense in my use case.

rohandubal:
Hello @michaeleconomy 

Thanks for reporting this. I will take the part of adding `clearStore` as an API as a feature request to the team.

Regarding your use-case, you can achieve this by scoping the cache to individual users and delete the database file once a user signs out. 

Example: https://github.com/aws-samples/aws-mobile-appsync-events-starter-ios/blob/master/EventsApp/AppDelegate.swift#L20 Here we are explicitly setting the database path. We can delete the file and create a new one for every user.

Hope that helps.

Thanks,
Rohan

joelmbell:
+1 on this feature request

larryonoff:
+1

shawnmmatthews:
+1

thegreatgabsby:
+1

larryonoff:
PR #141 is already in master!

thegreatgabsby:
actually ye, I managed to clear the store. cheers

palpatim:
This is released on 2.9.2. Please let us know if you have any issues.

PR #131, opened 2018-12-12
Bump Apollo to 0.9.4 except GraphQLOperation model

larryonoff:
*Issue #, if available:* #36

*Description of changes:*

* implement clear() for AWSSQLLiteNormalizedCache
* update Apollo to 0.9.4 except GraphQLOperation model

`GraphQLOperation` was not update since it's not compatible with the model that generates aws-codegen


palpatim:
Is the Apollo upgrade necessary for the `clear`? If not, are there particular features or bug fixes in the newer version of Apollo that makes the upgrade a priority?

larryonoff:
@palpatim 

> Is the Apollo upgrade necessary for the clear?

Yes. Because Apollo upgrade contains this feature.

zigfreid5:
Looking forward to this making it into the next release 🤞🏻

palpatim:
The primary reason for the change is the "clear" feature, and @larryonoff  has already done this on #141, which has a much reduced scope from this change, so I'm going to close this PR.

However, the core team would like to get back on an unforked version of Apollo if possible. So, within the next couple of months, we will  evaluate the changes we made to the Apollo code and see if we need to maintain the forked repository locally. As part of this, we'll see if we're able to isolate the changes into an easy-to-consume set of files, and add sufficient test coverage around our customized behavior that we are comfortable that upgrades won't cause any regressions to functionality we require for AppSync.

Issue #124, opened 2018-12-06
AWSAppSyncSubscriptionWatcher.connectedCallbackDelegate() Crash


larryonoff:
**Describe the bug**
We have the following crash report.

```
Crashed: com.apple.root.user-initiated-qos
0  libdispatch.dylib              0x18eb998f0 dispatch_group_leave$VARIANT$armv81 + 124
1  AWSAppSync                     0x1033b86a8 AWSAppSyncSubscriptionWatcher.connectedCallbackDelegate() (AWSAppSyncSubscriptionWatcher.swift:141)
2  AWSAppSync                     0x1033b86f4 @objc AWSAppSyncSubscriptionWatcher.connectedCallbackDelegate() (<compiler-generated>)
3  AWSAppSync                     0x10338bd74 $S10AWSAppSync03AppB10MQTTClientC23connectionStatusChanged_6clientySo16AWSIoTMQTTStatusV_So0I11TMQTTClientCyyXlyXlGtFyycfU_ (AppSyncMQTTClient.swift:43)
4  AWSAppSync                     0x103382e88 $SIeg_IeyB_TR (<compiler-generated>)
5  libdispatch.dylib              0x18ebc96c8 _dispatch_call_block_and_release + 24
6  libdispatch.dylib              0x18ebca484 _dispatch_client_callout + 16
7  libdispatch.dylib              0x18ebacc24 _dispatch_root_queue_drain + 640
8  libdispatch.dylib              0x18ebad35c _dispatch_worker_thread2 + 116
9  libsystem_pthread.dylib        0x18edac17c _pthread_wqthread + 472
10 libsystem_pthread.dylib        0x18edaecec start_wqthread + 4
```

```
AppSync.DeltaSyncOperationQueue.l9xfh4V0yae6HoiuGcCxbtWuAr1wmO1nKLhS1x7vuKo= (QOS: UNSPECIFIED)
0  libsystem_kernel.dylib         0x18ed279fc __ulock_wait + 8
1  libdispatch.dylib              0x18eb98e30 _dispatch_ulock_wait + 56
2  libdispatch.dylib              0x18eb99818 _dispatch_group_wait_slow + 40
3  AWSAppSync                     0x10339c8fc AppSyncSubscriptionWithSync.startSubscription() (AppSyncSubscriptionWithSync.swift:330)
[...]

PR #132, opened 2018-12-13
Refactored AppSyncSubscriptionWithSync

palpatim:
- Fix data races in AppSyncSubscriptionWithSync (Implicated in #124)
- Removed unused properties
- Updated project organization
- Added tests for refactored units

*Issue #, if available:*

Implicated in #124, but there are more and deeper races that we need to continue to work with.

*Description of changes:*


larryonoff:
I would like to add some recommendation on using Swift. It's marking all classes (especially internal) by default as `final`. This's because it's increases performance, see https://developer.apple.com/swift/blog/?id=27

AWSAppSyncClient/Internal/AppSyncSubscriptionWithSync.swift
```diff
@@ -0,0 +1,596 @@
...
+
+    // MARK: - Base query handling
+
+    /// Load base query results from cache, to quickly provide callers with the most recent results
+    private func initializeBaseQueryResultsFromCache() {
+        guard let appSyncClient = appSyncClient else {
+            return
+        }
+
+        let dispatchGroup = DispatchGroup()
```
larryonoff: I think that `DispatchGroup` has slightly different purpose, i.e. when you need to sync several async operations, but here you have only one async operation. So DispatchSemaphore suits better for this case.

AWSAppSyncClient/Internal/AppSyncSubscriptionWithSync.swift
```diff
@@ -0,0 +1,596 @@
...
+        dispatchGroup.wait()
+    }
+
+    /// Load base query results from service, forcing a bypass of the local cache
+    private func runBaseQueryFromService() -> Bool {
+        guard let appSyncClient = appSyncClient else {
+            return false
+        }
+
+        var success: Bool = true
```
rohandubal: I think we should initialize success to false. In case things go wrong at any place, we would want to return false.

AWSAppSyncClient/Internal/AppSyncSubscriptionWithSync.swift
```diff
@@ -0,0 +1,596 @@
...
+
+            defer {
+                semaphore.signal()
+            }
+
+            guard let self = self else {
+                return
+            }
+
+            // call customer if successful or error
```
rohandubal: nitpick: `customer` -> `developer`

AWSAppSyncClient/Internal/AppSyncSubscriptionWithSync.swift
rohandubal: This might be a good suggestion to also align with other similar implementation in our sync implementation

AWSAppSyncClient/Internal/AppSyncSubscriptionWithSync.swift
```diff
@@ -0,0 +1,596 @@
...
+                success = true
+                semaphore.signal()
+            }
+        }
+
+        let resultHandler: SubscriptionResultHandler<Subscription> = { [weak self] (result, transaction, error) in
+            // `subscribeWithConnectCallback` invokes `resultHandler` with an error if it encounters an error
+            // during the connect phase. Handle that here by updating the success flag and signalling the
+            // semaphore to end the method.
+            if error as? AWSAppSyncSubscriptionError != nil {
```
rohandubal: I think we should check for any error (not just subscription error)

AWSAppSyncClient/Internal/AppSyncSubscriptionWithSync.swift
larryonoff: Thank you. This's I forgot to mention.

AWSAppSyncClient/Internal/AppSyncSubscriptionWithSync.swift
larryonoff: Another idea is to introduce `internal enum` for such cases.

AWSAppSyncClient/Internal/AppSyncSubscriptionWithSync.swift
palpatim: Thanks, I neglected to replace this DispatchGroup with a semaphore, which I did elsewhere. Will change.

AWSAppSyncClient/Internal/SyncConfiguration.swift
```diff
@@ -0,0 +1,27 @@
...
+    internal var syncIntervalInSeconds: Int {
+        return seconds
+    }
+
+    public init(baseRefreshIntervalInSeconds: Int) {
+        self.seconds = baseRefreshIntervalInSeconds
+    }
+
+    // utility for setting default sync to 1 day
+    public class func defaultSyncConfiguration() -> SyncConfiguration {
```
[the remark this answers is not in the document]
palpatim: Good callout, will make that change here and add it to our (as yet unpublished) coding style/best practice guide.

palpatim:
> I would like to add some recommendation on using Swift. It's marking all classes (especially internal) by default as `final`. This's because it's increases performance, see https://developer.apple.com/swift/blog/?id=27

Good callout, thanks.

AWSAppSyncClient/Internal/AppSyncSubscriptionWithSync.swift
palpatim: Thanks, I'll look at this again

AWSAppSyncClient/Internal/AppSyncSubscriptionWithSync.swift
larryonoff: I think it may be good to have 'retryCount' limit. So if there're `errors.count` > `retryCount`. Then restart of subscription won't happen anymore and e.g. should be done manually.

AWSAppSyncClient/Internal/SubscriptionMessageQueue.swift
```diff
@@ -0,0 +1,110 @@
...
+    /// Note: this method must only be invoked from a `syncQueue.sync` block, as it mutates the messages queue.
+    private func drainQueue() {
+        for queueItem in messagesQueue {
+            deliver(item: queueItem)
+        }
+        messagesQueue = []
+    }
+
+    /// Asynchronously invokes the resultHandler for `item` on the global queue
+    private func deliver(item: QueueItem, transaction: ApolloStore.ReadWriteTransaction? = nil) {
```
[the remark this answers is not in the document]
palpatim: Thanks for the callout, will do.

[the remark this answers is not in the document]
larryonoff: just as sample

1. https://developer.apple.com/documentation/swift/array/1538872-append
2. https://developer.apple.com/documentation/swift/array/2885400-insert

AWSAppSyncClient/Internal/SyncConfiguration.swift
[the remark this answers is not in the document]
larryonoff: In the case of using `default` as singleton you just can do the following
```
public static let `default`: SyncConfiguration {
//  ... initialisation
}()
```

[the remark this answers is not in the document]
larryonoff: I also think that value type i.e. `struct` is better for SyncConfiguration.

AWSAppSyncClient/Internal/AppSyncSubscriptionWithSync.swift
palpatim: (However, I'm not going to refactor all the other usages of DispatchGroup as part of this PR.)

AWSAppSyncClient/Internal/AppSyncSubscriptionWithSync.swift
palpatim: Simplified the call and obviated the comment :)

AWSAppSyncClient/Internal/AppSyncSubscriptionWithSync.swift
palpatim: > I think we should check for any error (not just subscription error)

Since we have a `success == nil` check, that makes sense.

> I think it may be good to have 'retryCount' limit. So if there're `errors.count` > `retryCount`. Then restart of subscription won't happen anymore and e.g. should be done manually.

I'd rather do that as part of our TODO to handle subscription setup separately from sync. Currently, `startSubscription` is only ever invoked from `performSync`, because we don't have good lifecycle and network state awareness on subscriptions. Once we tease apart subscription handling from sync handling, having separately-configured and managed retry strategies for sync and subscription setup will be appropriate. 

Regarding "should be done manually": Since the overall consistency of the system is pretty tightly interwoven, and depends on quite a few moving parts, what would your view of a recovery mechanism look like?

larryonoff: I think that the best solution is giving API user to control retry count and when to retry, e.g. in case of FRP we can have something like
```swift
appSyncClient
   .sync(...)
   .retry(count: 5)
   .flatMapError(...)
   .subscribe()
```

But this's not what AppSync API needs to introduce, it's already Rx. But we're not in FRP / Rx world.

So here's my opinion:
 1. `AppSyncSubscriptionWithSync` should start / stop subscriptions only when reachability events occur.
 2. start / stop count should be a part `SyncConfiguration` with default value, like `let reachabilityRetryCount: Int = .max`.
 2. `AppSyncSubscriptionWithSync` should check reachability and wait (not restart as now) while host is reachable according to the current client settings.
 3. Some error should be forwarded to outer layer when `reachabilityRetryCount` reached.
 4. Other than Reachability errors should be forwarded to the outer layer and user should decide what to do.
 5. AWSAppSync errors should be revised, since it's wrapped strings in general. So that it's difficult and unsafe to match concrete error.

AWSAppSyncClient/Internal/SubscriptionWithSyncConfiguration.swift
```diff
@@ -0,0 +1,29 @@
...
+// http://aws.amazon.com/apache2.0
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+//
+
+/// Configuration values for an AppSyncSubscriptionWithSync instance.
+public struct SubscriptionWithSyncConfiguration {
```
larryonoff: I think this not clear naming. I thought that it's new `Subscription` class at once, but it's configuration of `SubscriptionWithSync`.

AWSAppSyncClient/Internal/SubscriptionWithSyncConfiguration.swift
```diff
@@ -0,0 +1,29 @@
...
+    /// The interval, in whole seconds, at which the subscription will be refreshed using the `deltaQuery`. If more time has
+    /// elapsed since the last sync, then local data will be refreshed using `baseQuery` instead.
+    let syncIntervalInSeconds: Int
+
+    /// Creates a new SubscriptionWithSyncConfiguration with the specified sync interval.
+    ///
+    /// - Parameters:
+    ///   - syncIntervalInSeconds: The sync interval. Defaults to one day (86,400 seconds)
+    init(syncIntervalInSeconds: Int = 86_400) {
```
rohandubal: Won't this initializer be inaccessible to developers? Is that the intention?

AWSAppSyncClient/Internal/SubscriptionWithSyncConfiguration.swift
rohandubal: I think we should remove the word `Subscription` from this struct name. We don't necessarily use subscription for the sync API. Maybe something like `SyncIntervalConfiguration`?

larryonoff: I think that the old name is much clearer.

AWSAppSyncClient/AWSAppSyncClient.swift
```diff
@@ -918,13 +930,88 @@ public class AWSAppSyncClient {
...
+    ///   - baseQueryResultHandler: Closure that is called when base query results are available or when an error occurs. Every time a sync operation is called, a fetch for the baseQuery from the cache will be done first before initiating any other operations.
+    ///   - subscription: The subscription which will provide real time updates.
+    ///   - subscriptionResultHandler: Closure that is called when a real time update is available or when an error occurs.
+    ///   - deltaQuery: The delta query which fetches data starting from the `lastSync` time.
+    ///   - deltaQueryResultHandler: Closure that is called when delta query executes.
+    ///   - callbackQueue: An optional queue on which sync callbacks will be invoked. Defaults to the main queue.
+    ///   - subscriptionWithSyncConfiguration: The sync configuration where the baseQuery sync interval can be specified. (Defaults to 24 hours.)
+    /// - Returns: An object that can be used to cancel the sync operation.
+    public func sync<BaseQuery: GraphQLQuery, Subscription: GraphQLSubscription, DeltaQuery: GraphQLQuery>(baseQuery: BaseQuery,
```
larryonoff: I think that parameter `subscriptionWithSyncConfiguration` can just be named `configuration` or `config`. It's already domain specific to the function `sync`.

AWSAppSyncClient/AWSAppSyncClient.swift
```diff
@@ -918,13 +930,88 @@ public class AWSAppSyncClient {
...
+    ///   - baseQueryResultHandler: Closure that is called when base query results are available or when an error occurs. Every time a sync operation is called, a fetch for the baseQuery from the cache will be done first before initiating any other operations.
+    ///   - deltaQuery: The delta query which fetches data starting from the `lastSync` time. Defaults to a no-op query.
+    ///   - deltaQueryResultHandler: Closure that is called when delta query executes. Defaults to a no-op handler.
+    ///   - callbackQueue: An optional queue on which sync callbacks will be invoked. Defaults to the main queue.
+    ///   - subscriptionWithSyncConfiguration: The sync configuration where the baseQuery sync interval can be specified. (Defaults to 24 hours.)
+    /// - Returns: An object that can be used to cancel the sync operation.
+    public func sync<BaseQuery: GraphQLQuery, DeltaQuery: GraphQLQuery>(baseQuery: BaseQuery,
+                                                                        baseQueryResultHandler: @escaping OperationResultHandler<BaseQuery>,
+                                                                        deltaQuery: DeltaQuery? = nil,
```
rohandubal: marking `deltaQuery? = nil` will probably cause compilation error if un-specified.

AWSAppSyncClient/AWSAppSyncClient.swift
```diff
@@ -918,13 +930,88 @@ public class AWSAppSyncClient {
...
+                                                                        deltaQueryResultHandler: DeltaQueryResultHandler<DeltaQuery>? = nil,
```
rohandubal: same as above

AWSAppSyncClient/AWSAppSyncClient.swift
```diff
@@ -918,13 +930,88 @@ public class AWSAppSyncClient {
...
+                                                                        callbackQueue: DispatchQueue = DispatchQueue.main,
+                                                                        subscriptionWithSyncConfiguration: SubscriptionWithSyncConfiguration? = nil) -> Cancellable {
```
rohandubal: can we rather use default parameter of SubscriptionWithSyncConfiguration()?

```swift
subscriptionWithSyncConfiguration:  SubscriptionWithSyncConfiguration = SubscriptionWithSyncConfiguration()
```

AWSAppSyncClient/Internal/SubscriptionWithSyncConfiguration.swift
larryonoff: I think `SyncIntervalConfiguration` can be used not only for sync interval parameter, so from my opinion `SyncConfiguration` is the best.

AWSAppSyncClient/Internal/SubscriptionWithSyncConfiguration.swift
larryonoff: this won't be accessible for developers, since it's `internal` by default.

AWSAppSyncClient/AWSAppSyncClient.swift
larryonoff: it should not, since default value is `nil`.

rohandubal: The generic type would not be able to inferred causing the error I think.

AWSAppSyncClient/AWSAppSyncClient.swift
larryonoff: I think default value that is not `nil` is clearer. `nil` may be more memory effective, but sync config is created in the function body below. So it doesn't matter in this case.

AWSAppSyncClient/AWSAppSyncClient.swift
larryonoff: I'm not sure about this. It may be compiler bug if nil doesn't work. Did you check it?

AWSAppSyncClient/Internal/AppSyncSubscriptionWithSync.swift
palpatim: I've got some thoughts on the overall pattern, but let's move the retry handling discussion off to another thread, so we don't pollute the intent of this PR.

AWSAppSyncClient/Internal/SubscriptionWithSyncConfiguration.swift
palpatim: Thanks, re-looking at this with a day of distance, I agree that the [new name is a bit crap](https://martinfowler.com/bliki/TwoHardThings.html). :) However, as much as I like the existing name, I don't want to change it to a `struct` since that would break anybody relying on the existing class semantics of that type. (It's unlikely, but I don't want to break an API contract without warning.) Let me consider this.

palpatim: > I think `SyncIntervalConfiguration` can be used not only for sync interval parameter, so from my opinion `SyncConfiguration` is the best.

That's my concern as well--if we're going to the bother of having a configuration type in the first place, it should probably be a bit broader than just an interval configuration.

AWSAppSyncClient/AWSAppSyncClient.swift
palpatim: Also, you can't use non-statically-created values as defaults--that was my first thought too :)

AWSAppSyncClient/AWSAppSyncClient.swift
palpatim: It definitely compiled fine for me locally. I'll try it in a few other contexts.

AWSAppSyncClient/Internal/SubscriptionWithSyncConfiguration.swift
palpatim: OK, @rohandubal convinced me--I'll convert `SyncConfiguration` to a `struct` rather than deprecating the previous type. That will make this PR a breaking change, but we expect the actual impact on developers to be very small since this is a relatively new API, and it would be a pretty uncommon pattern to be relying on class semantics for this particular type.

palpatim: Resolved as noted elsewhere: `SyncConfiguration` is now a `struct`, and changed the initializer parameter and default pattern.

AWSAppSyncClient/AWSAppSyncClient.swift
palpatim: It compiled OK for me, but in the interest of making the API very explicit about the supported use cases, I've made these required instead of optional. The three `sync` flavors are:
1. base only
1. base plus delta
1. base plus delta plus subscription

By not making the `delta` optional, it will make it clearer to callers that these are the three supported use cases (at the expense of a larger `sync` method family).

AWSAppSyncClient/AWSAppSyncClient.swift
palpatim: Revised and included `SyncConfiguration()` as the default here and throughout. I had mistakenly thought that using an instance constructor wasn't allowed for defaults; the problem was actually that the constructor was `internal` rather than `public`, as @rohandubal noted elsewhere.

AWSAppSyncClient/Internal/SubscriptionWithSyncConfiguration.swift
palpatim: Made initializer public, which also solved my problem using `SyncConfiguration()` as a default value for method parameters.

AWSAppSyncClient/Internal/SubscriptionWithSyncConfiguration.swift
larryonoff: I also think that it's not a big deal replacing `struct` with `class`.

1. I'm not sure that there're a lot users who use AWSAppSync iOS SDK and the latest feature `sync`.
2. There weren't Swift versions that don't have breaking changes. So I think this isn't new for Swift developers :)

Issue #122, opened 2018-12-04
S3 upload doesn't work


nourahassan:
I followed the documentation here https://aws-amplify.github.io/docs/ios/storage#usage-with-graphql-apis-complex-objects

First the code generated by `aws-appsync-codegen` has a compilation issue, it adds an extension to S3Object and  compiler complains it is not configured. There is no struct S3Object only S3ObjectInput.

I commented this code:

```
extension S3Object: AWSS3ObjectProtocol {
  public func getBucketName() -> String {
      return bucket
  }

  public func getKeyName() -> String {
      return key
  }

  public func getRegion() -> String {
      return region
  }
}
```

Then when I create my Post with the file, it saves the post to dynamoDB but not the file to S3. And there is no single error?!!

And I even added a breakpoint to created extension for AWSTransferUtilities created by the codegen and it is never called??
 
 - AppSync SDK Version: 2.7.0
 - Dependency Manager: Carthage
 - Swift Version : 4.1
 - Device:  Simulator
 - iOS Version: 12.1

rohandubal:
Hello @nourahassan 

Sorry that you are having issues using the SDK.

Could you please provide with the schema which you are using? I will try to put up a small sample which demonstrates this feature. You should not have to delete any generated code. 

Thanks,
Rohan

kjlis:
I'm experiencing exactly the same problem. In the beginning I was following [the amplify guide](https://aws-amplify.github.io/docs/cli/graphql#s3-objects) which mentions only adding `S3Object` type:
```
type S3Object {
  bucket: String!
  region: String!
  key: String!
}
```
After running `amplify push` GraphQL API gets updated, the auto-upload didn't work, so I followed the [guide from Storage section](https://aws-amplify.github.io/docs/ios/storage#usage-with-graphql-apis-complex-objects) and added 
```
type Item @model @auth(rules: [{ allow: owner }) 
{
  id: ID!
  owner: String!
  itemName: String!
  file: S3Object
}
  type S3Object {
          bucket: String!
          key: String!
          region: String!
  }
  input S3ObjectInput {
          bucket: String!
          key: String!
          region: String!
          localUri: String!
          mimeType: String!
  }
```
to `backend/api/~apiname~/schema.graphql` and regenerated the code by running:
```
amplify push &&
aws-appsync-codegen generate graphql/*.graphql --schema graphql/schema.json --output API.swift --add-s3-wrapper
```
[...]

PR #136, opened 2018-12-22
Client config refactoring and tests

palpatim:
This PR is in preparation for adding tests & a fix for #122 (but does not actually address that fix yet). In order to properly test upload & queueing, we need better testability and control over client configuration, which this PR lays the groundwork for.

- Added IDETemplateMacros to propagate copyright header for new files
- Added unit tests for S3 object uploads
- Reorganized tests
- Reorganized some internal classes
- Refactored ClientConfiguration & ClientInfo
- Added implementation for AWSCognitoUserPoolsAuthProviderAsync and
  AWSOIDCAuthProviderAsync test classes to prevent intermittent failures
- Updated the Events API to include S3 file object, again in preparation for #122.

Of special note is the refactoring around the AWSAppSyncClientConfiguration init methods. I consolidated them into two pairs of methods.
- One pair that allows callers to specify their own networkTransport
- One pair that lets the init method derive the networkTransport to create, by inspecting the incoming auth providers.

Each pair has a convenience initializer that takes an AWSAppSyncClientInfo object, and a designated initializer that takes specified values.

This refactoring removes some method signatures, but I don't *think* it's actually a breaking change, since the revised signatures simply do the same thing with optional / default values. There's a whole lotta `nil` in the method signature, but it'll take a larger refactoring to tease apart the auth configurations from the AWSAppSyncClientConfiguration class.

Finally as part of this, I added tests that cover the config (100%) and info (95%) classes.

*Issue #, if available:*
N/A, although this might include some of the suggested changes from @larryonoff  over a few of their outstanding PRs. I will review & attribute accordingly next week.

*Description of changes:*


AWSAppSyncClient/Internal/AWSAppSyncClientInfo.swift
```diff
@@ -0,0 +1,73 @@
...
+        static let root = "AppSync"
+        static let apiURL = "ApiUrl"
+        static let region = "Region"
+        static let authMode = "AuthMode"
+        static let apiKey = "ApiKey"
+    }
+
+    public let apiUrl: String
+    public let region: String
+    public let authType: String
```
larryonoff: I suggest here use `AuthType` instead of just `String`. And also rename `AuthType` into `AWSAppSyncAuthType`.

AWSAppSyncClient/Internal/AWSAppSyncClientInfo.swift
```diff
@@ -0,0 +1,73 @@
...
+    var apiUrl: String { get }
+    var region: String { get }
+    var authType: String { get }
+    var apiKey: String { get }
+}
+
+/**
+ * Configuration for AWSAppSyncClient as read from `awsconfiguration.json`
+ */
+public class AWSAppSyncClientInfo: AWSAppSyncClientInfoProviding {
```
larryonoff: It would be great if you take changes from https://github.com/awslabs/aws-mobile-appsync-sdk-ios/blob/25935a273ab7092d6a49c1868d9099379033daab/AWSAppSyncClient/AWSAppSyncClientInfo.swift.

This removed force unwraps, which are really bad thing.

palpatim: Thanks for the suggestions, I will incorporate that.

AWSAppSyncClient/Internal/AWSAppSyncClientInfo.swift
```diff
@@ -0,0 +1,73 @@
...
+    private struct ConfigFileKeys {
+        static let root = "AppSync"
+        static let apiURL = "ApiUrl"
+        static let region = "Region"
+        static let authMode = "AuthMode"
+        static let apiKey = "ApiKey"
+    }
+
+    public let apiUrl: String
+    public let region: String
```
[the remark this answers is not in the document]
palpatim: Since this is a public class (which I mistakenly moved into the "Internal" directory and will restore), I don't really want to change the API without warning. However, I will rework the protocol to use strong types, provide a new instantiating type, and deprecate the `AWSAppSyncClientInfo` class (which doesn't really need to be a class anyway).

AWSAppSyncClient/Internal/AWSAppSyncClientInfo.swift
palpatim: I agree with this change: Since `AuthType` was an internal type, renaming without a breaking change is fine. Since I'm promoting it to public visibility by making it the type of one of the configurations, namespacing it is also appropriate.

palpatim: Er, I meant to say, "I agree with this suggestion", or more succinctly, "Thanks, will do!" :)

AWSAppSyncClient/AWSAppSyncServiceConfig.swift
```diff
@@ -0,0 +1,111 @@
...
+
+        apiKey = configForKey[AWSConfigurationFile.Keys.apiKey] as? String
+
+        guard authType != AWSAppSyncAuthType.apiKey || !apiKey.isEmpty else {
+            throw AWSAppSyncServiceConfigError.invalidAPIKey
+        }
+    }
+}
+
+public enum AWSAppSyncServiceConfigError: String, Error, LocalizedError {
```
larryonoff: Just to consider. I usually extend classes separately from the main declaration, i.e.

```
// MARK: - LocalizedError

extension AWSAppSyncServiceConfigError: LocalizedError {
...
}

Comment (palpatim) [reply AWSAppSyncClient/AWSAppSyncServiceConfig.swift]: Thanks for the note. I'm of two minds about this, and have done both ways (possibly in this very PR) by including the extension in the main body of the type and also doing a separate extension. As many others have noted, Swift doesn't allow you to define storage in an extension, so protocol conformance might be broken into multiple sections (base class for properties to support the protocol; an extension to define the behavior). However, that's probably the right thing to do anyway.

I'll consider this with the team and we'll make an official decision on the style we want to adopt for the project. For now, I'll move the conformance into a separate block as you suggest--I agree it keeps code more readable overall.

Comment (larryonoff) [reply AWSAppSyncClient/AWSAppSyncServiceConfig.swift]: Thanks for accepting note.

I think that this not only for readability. This approach also allows to throw away some obsolete or unnecessary code easily.

As far as I remember Apple Swift team also suggests this approach. But I cannot find link.

There’s some limitations, e.g. Codable / Hashable / Equatable auto synthesizing. You cannot extend class automatically with separate extension. But I personally prefer  separate extensions with some exceptions.

File: AWSAppSyncClient/AWSAppSyncServiceConfig.swift [hunk 5]
```diff
@@ -0,0 +1,111 @@
...
+    /// The authentication method used to access the API. If this value is `.apiKey`, then the config must also provide a value
+    /// for `apiKey`
+    var authType: AWSAppSyncAuthType { get }
+
+    /// If `authType` is `.apiKey`, this value must provided
+    var apiKey: String? { get }
+}
+
+/// Client-side configurations of an AWSAppSync service instance
+public struct AWSAppSyncServiceConfig {
```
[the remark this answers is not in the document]
palpatim: Good callout, will do

AWSAppSyncClient/AWSAppSyncServiceConfig.swift
larryonoff: Not Apple, but good team https://github.com/raywenderlich/swift-style-guide/blob/master/README.markdown#protocol-conformance

palpatim:
Test failures are due to intermittent service-side throttling. Merging.

PR #147, opened 2019-01-08
Network status refactoring & tests, MutationQueue tests

palpatim:
*Issue #, if available:*
#109, #139

*Description of changes:*
- Adds mutation queue unit tests supporting #109 
- Refactors network state handling #139
- Adds unit tests for network state handling

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Issue #139, opened 2018-12-31
Wrap Reachability in a protocol to aid testability


palpatim:
**Is your feature request related to a problem? Please describe.**

We currently inject Reachability directly into consuming types. We should wrap it in a protocol to aid testability. That should lay the groundwork for better test coverage of mutation queue behavior when the network stops/resumes.

**Describe the solution you'd like**

Extract the Reachability functions we're currently using into a protocol; and conform Reachability to it. Replace directly-instantiated Reachability client with init params.

**Describe alternatives you've considered**

None. Reachability is a popular framework that provides useful utilities. It doesn't make sense to roll our own, and a protocol wrapper is a cheap enhancement.

**Additional context**

N/A

larryonoff:
One more thing I would like to suggest is moving Reachability inside the framework, not as dependency of AppSync. This will exclude conflicts when someone uses newer or older version of dependency, so that AppSync update won't be possible until updating dependency. 

It can be done via git submodules or Carthage.

larryonoff:
@palpatim I see several options here.

**1. CocoaPods dependency or Carthage dependency**

I think this's popular option. The only disadvantage I see is dependency versions conflict if someone uses newer or older version of the dependency that AppSync uses.

**2. Internal dependency as a source, NOT as framework**

The advantage that it will eliminate versions conflict.

The disadvantage that it's harder to manage, e.g. git submodules, Carthage via source code.

**3. Internal dependency as framework**

This's the worse scenario, because
1. Swift ABI not stable.
2. Dependency is bundled and hidden inside the framework.


I think that options 1 and don't hide what dependencies used via dependencies list or source code.

palpatim:
Hi @larryonoff,

I agree with your assessment, and choosing what you're outlining as option 1, with a modification that makes it behave like Option 2: I'm currently testing adding all of our Pods to source control (#150), which would remove the need for developers who consume this library via Carthage or source to install CocoaPods and run a `pod update` before building.

* Carthage developers who consume binaries (that is, who are using the same version of Swift we use to build the SDK) will be unaffected.
* Developers who consume from source (either `carthage update --no-use-binaries`, using a git submodule, etc) should be able to build without a `pod install`

Dependencies will continue to be documented within the Cartfile and Podfile (except for Apollo as we noted above), but won't require action on the developer's part to update.

[...]

PR #148, opened 2019-01-09
Palpatim/mutation retry test

palpatim:
*Issue #, if available:*

This includes changes from #147.

*Description of changes:*

Tests & minor cleanup

- Added tests for cancelled mutations and resuming dropped network, supporting #109 
- Moved some classes to internal
- Added retry handling unit tests


AWSAppSyncClient/Internal/AWSAppSyncRetryHandler.swift
```diff
@@ -76,7 +80,27 @@ final class AWSAppSyncRetryHandler {
...
     }
+
+    /// Returns the value of the "Retry-After" header as an Int, or nil if the value isn't present or cannot be converted to
+    /// an Int
+    ///
+    /// - Parameter response: The response to get the header from
+    /// - Returns: The value of the "Retry-After" header, or nil if not present or not convertable to Int
+    private static func getRetryAfterHeaderValue(from response: HTTPURLResponse) -> Int? {
+        let waitTime: Int?
+        switch response.allHeaderFields["Retry-After"] {
```
rohandubal: Not particular related to this change, but I think we should consider doing lowercase/ uppercase for this header to make sure we don't miss this information due to case.

palpatim: No need, the `allHeaderFields` dictionary keys are [case-insensitive](https://developer.apple.com/documentation/foundation/httpurlresponse/1417930-allheaderfields)

palpatim:
Since this includes the changes from #147, it completes issues #109 and #139

PR #155, opened 2019-01-17
Make AWSMutationCache thread safe

larryonoff:
*Issue #, if available:*

NA.

*Description of changes:*

We still having issues with restoring offline mutations, sometimes application just hangs. So I started migration `AWSMutationCache` into promises and I'm going to implement separate `DispatchQueue` for `AWSMutationCache`.


palpatim:
Thanks for the PR @larryonoff. I will review at my earliest opportunity. In order to merge, we'll need tests to reproduce the incorrect behavior. Thread safety is always tricky, so I'm considering options here--one might be to turn on TSan and run tests that add mutations from different queues.

larryonoff:
@palpatim this PR is not complete. I've just added Promises somewhere. I created PR earlier that implemented because wanted to provide the issue and some ideas of fixing it.

palpatim:
Hi @larryonoff,

I know this is still a work in progress, but I'm going to merge this into `palpatim/crash-resuming-from-background` as part of my work on #160. I've been debating about using a Promise-based solution to simplify our async programming model. I'm not 100% convinced that the Apollo Promises implementation is how I want to go, but as long as we limit their use to internal APIs, it seems like a relatively easy way to start standardizing our async code.

larryonoff:
@palpatim thanks for merging.

I don’t work on this functionality at the moment and not sure that I will. We decided to move away from AppSync. So I’m not sure that I’ll continue with AppSync PRs.

I also dislike Apollo Promises solution. The best approach is having Promises in Swift Foundation, but it is not available at the moment. So the next that I like is ReactiveSwift (RxSwift is also good, but not for me), but RFP is a difficult paradigm. BTW I can share some extensions that I’m using with ReactiveSwift. I’m sure that there’re good solutions already implemented by others, so I don’t recommend to do in-home solution. I think it would be better to git-submodule something really good.

palpatim:
Sorry to hear you've decided not to use AppSync. If you'd like to provide feedback on why it is not suitable (or became unsuitable), either here or via private channels, we'd love to know more.

I like the idea of stream programming like RxSwift/ReactiveSwift, but it introduces enough complexity that we'll have to think very hard about adopting a fundamental change like that.

I agree, having Promises as a native language feature would be spectacular. At least we're getting `Result`s in Swift 5. :)

larryonoff:
@palpatim one of the main reasons is that AppSync has really critical issues, like deadlocks (new issue) or offline mutations didn't work (that was fixed already). We're really small team (with one iOS developer me). So we just cannot spend time on fixing / improving external tools that we're using.

> I agree, having Promises as a native language feature would be spectacular. At least we're getting Results in Swift 5. :)

This's absolutely! Looking forward to async / await next.

PR #156, opened 2019-01-17
Fix valid architectures

larryonoff:
*Issue #, if available:*

NA

*Description of changes:*

Currently for some reason valid architectures are only "armv7 arm64 armv7s", but for example the latest Xcode 10 supports `arm64e`.  I think that valid architectures should be default that Xcode sets if there're no any specific reasons.


palpatim:
Good callout, thanks. Testing locally.

palpatim:
Released on 2.9.2

Issue #76, opened 2018-09-23
Carthage build reachability framework is outdated


graylikeme:
**Describe the bug**
When building with Carthage the version of Reachability that is being downloaded is 3, but the sdk binary attempts to use functions from v4

`dyld: Symbol not found: _$S12ReachabilityAAC24allowsCellularConnectionSbvpWvd
  Referenced from: /private/var/containers/Bundle/Application/54737AA0-D181-4133-830C-E863D0AA335B/Pear-DEV.app/Frameworks/AWSAppSync.framework/AWSAppSync
  Expected in: /private/var/containers/Bundle/Application/54737AA0-D181-4133-830C-E863D0AA335B/Pear-DEV.app/Frameworks/Reachability.framework/Reachability
 in /private/var/containers/Bundle/Application/54737AA0-D181-4133-830C-E863D0AA335B/Pear-DEV.app/Frameworks/AWSAppSync.framework/AWSAppSync
`

**Environment(please complete the following information):**
 - AppSync SDK Version: [2.6.22]
 - Dependency Manager: [Carthage]
 - Swift Version : [4.2]

**Additional context**
The problem seems to be the version specified in Cartfile (specified to be compatible with v 0.4.0). Needs an update to the latest

pawelurb:
For people currently struggling with this issue, there's a workaround for this problem that you can apply by explicitly defining version of Reachability in your Cartfile:

```
github "awslabs/aws-mobile-appsync-sdk-ios" ~> 2.6.0
github "ashleymills/Reachability.swift" "v4.2.1"
```

Hint: if you're building with Swift 4.2 then [v4.3.0 has been released](https://github.com/ashleymills/Reachability.swift/releases), so you can update your Cartfile accordingly.

However, I also agree that this issue should be addressed in the library source files.

kiroskirin:
@pawelurb Still got error even doing as your workaround.

pawelurb:
@kiroskirin can you paste the content of your Cartfile please and let me know which version of Swift are you using?

peterfennema:
Same issue here, swift 4.2. The workaround of @pawelurb does not work. 
The v3 version of Reachability is downloaded, the 4.3.0. version is ignored.

Cartfile:

```
github "awslabs/aws-mobile-appsync-sdk-ios" == 2.6.22
github "ashleymills/Reachability.swift" == 4.3.0

```

pawelurb:
[...]

PR #158, opened 2019-01-21
Remove dependencies from cartfile

palpatim:
*Issue #, if available:*
#76

*Description of changes:*
Removes dependencies from `Cartfile` and points people to the Podfile.lock for a canonical list of downloadable dependencies.


palpatim:
After conversation with @rohandubal I'm withdrawing this CR, as it would break transitive dependencies for Carthage-built apps.

Issue #160, opened 2019-01-22
Crash when returning from background - SIGTRAP(TRAP_BRKPT)


adamup928:
**Describe the bug**
When returning from the background, we experience intermittent crashes related to the AppSync SDK.

**To Reproduce**
1.  Initialize an AppSync instance.
2.  Minimize the app (while not connected to a debugger).
3.  When the app returns to an active state, occasionally it will crash with the AWS AppSync module referenced in the stack trace.

**Expected behavior**
Returning from the background should not result in a crash.

**Environment(please complete the following information):**
 - SDK Version: AWS Core 2.8.4, AWSAppSync 2.9.1
 - Dependency Manager: Cocoapods
 - Swift Version : 4.2

**Device Information (please complete the following information):**
 - Device: Multiple (iPhone X, iPhone 7 Plus)
 - iOS Version: [e.g. iOS 11.4] 12.1 and 12.1.2
 - Specific to simulators: n/a

**Additional context**

Here is a symbolicated stack trace:

```Cause: Crash due to signal: SIGTRAP(TRAP_BRKPT) at 10bf4a1a4```

```Thread 16 [Crashed]:

0    libswiftCore.dylib                       0x10bf4a1a4     $Ss17_assertionFailure__4file4line5flagss5NeverOs12StaticStringV_SSAHSus6UInt32VtFTf4nxnnn_n + 164
1    libswiftCore.dylib                       0x10bf9fc48     swift_unexpectedError + 280
2    SQLite                                   0x10a334464     (null)
3    SQLite                                   0x10a33eea0     (null)
4    libswiftCore.dylib                       0x10bf2e320     $Ss21_ClosureBasedIteratorV4nextxSgyF + 48
5    libswiftCore.dylib                       0x10bf2e4ac     $Ss12_IteratorBoxC4next7ElementQzSgyF + 36
6    libswiftCore.dylib                       0x10bdf0050     $Ss11AnyIteratorV4nextxSgyF + 28
7    libswiftCore.dylib                       0x10bd5e9c8     $SSTsE3mapySayqd__Gqd__7ElementQzKXEKlF + 816
8    libswiftCore.dylib                       0x10bf39390     $Ss21_ClosureBasedSequenceVyxGSTsST3mapySayqd__Gqd__7ElementQzKXEKlFTW + 84
[...]

PR #171, opened 2019-02-05
Palpatim/crash resuming from background

palpatim:
*Issue #, if available:*
#160 

*Description of changes:*

* Splits databases into separate files to reduce contention. Provides migration utility from old-style to new-style cache.
* Wires Promises into MutationQueue flow to make async model easier to follow. Lays groundwork for future Promises integration.


AWSAppSyncClient/Internal/AWSMutationCache.swift
```diff
@@ -0,0 +1,153 @@
...
+public final class AWSMutationCache {
+
+    private let db: Connection
+    private let mutationRecords = Table("mutation_records")
+    private let id = Expression<Int64>("_id")
+    private let recordIdentifier = Expression<CacheKey>("recordIdentifier")
+    private let data = Expression<Data>("data")
+    private let recordState = Expression<String>("recordState")
+    private let timestamp = Expression<Date>("timestamp")
+    private let s3Bucket = Expression<String?>("s3Bucket")
```
rohandubal: With some customer asks about supporting more than 1 s3 file/ complex object per mutation, do you think it is a good idea to use this opportunity to change the schema to hold an array of s3 objects?

AWSAppSyncClient/Internal/SQLiteSerialization.swift
```diff
@@ -0,0 +1,66 @@
...
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+//
+
+import Foundation
+import SQLite
+
+final class SQLiteSerialization {
```
rohandubal: nice!

AWSAppSyncClient/AWSAppSyncCache.swift
```diff
@@ -0,0 +1,93 @@
...
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+//
+
+import Foundation
+
+/// Errors thrown from the Queries Cache
+public enum AWSAppSyncQueriesCacheError: Error {
```
rohandubal: Is there a way in which `AWSAppSyncQueriesCacheError` can inherit from `AWSSQLLiteNormalizedCacheError ` so that the customer experience can be slightly better? We can still deprecate `AWSSQLLiteNormalizedCacheError` and get rid of it eventually.

AWSAppSyncClient/AWSAppSyncCache.swift
```diff
@@ -0,0 +1,93 @@
...
+    /// Creates a cache configuration with individually-specified cache paths. If a path is nil, the cache will
+    /// be created in-memory. If specified, the directory portion of the file path must exist, and be writable by the
+    /// hosting app. No attempt is made to validate the specified file paths until AppSync initializes the related
+    /// cache at the specified path.
+    ///
+    /// - Parameters:
+    ///   - offlineMutations: The file path to create or connect to the cache for the offline mutation queue.
+    ///   - queries: The file path to create or connect to the cache for the local query cache.
+    ///   - subscriptionMetadataCache: The file path to create or connect to the cache for subscription metadata.
+    init(offlineMutations: URL?, queries: URL?, subscriptionMetadataCache: URL?) {
```
rohandubal: Should this be marked `public`?

AWSAppSyncClient/Internal/AWSMutationCache.swift
palpatim: I'd like to put that in a separate change, since this PR is a) targeted toward refactoring database handling; and b) quite large enough. :)

AWSAppSyncClient/AWSAppSyncCache.swift
palpatim: Great catch, thanks.

AWSAppSyncClient/AWSAppSyncCache.swift
palpatim: I considered some options but couldn't find a way to make this a non-breaking change. Value types like Enums can't inherit in Swift, and any change to the throwing type is going to potentially impact customers who are catching that specifically.

The three options I considered:
1. Leave it as-is
2. Remove `AWSSQLLiteNormalizedCacheError` in favor of the new error type
3. Deprecate `AWSSQLLiteNormalizedCacheError` in favor of the new error type

I don't favor Option 1. `AWSSQLLiteNormalizedCacheError` as a type name has some semantic confusion (What's a normalized cache? Why is SQLLite specified? Why is SQLite misspelled? :) )

I went back & forth between Option 2 & Option 3 and only settled on deprecation because that's our traditional way of doing things. I could be convinced to do Option 2--just removing it altogether. Deprecating means that customers who are actually taking meaningful action based on the error type will see a warning, but won't be forced to deal with the potential change in behavior. However, for customers who aren't doing anything critical (e.g. they're adding or suppressing logging based on the error type), it's an unexpected change to remove it.

A deprecation with a FixIt seemed the best compromise.

rohandubal: Sounds good, thanks for the detailed explanation!

Issue #193, opened 2019-03-05
AppSync subscriptions auto-reconnect while advising clients to reconnect


palpatim:
if the AWSIoTMQTTClient receives a web socket disconnect that is not the result of a user disconnect request, it will notify its delegate of the error, but also attempt a reconnect on its own.

@cbommas advises that he fixed this problem on Android by suppressing the auto-reconnect behavior at the MQTT low level client layer, in favor of managing AppSync subscription reconnect logic at the AppSync client layer.

The preferred fix would be to offer a flag in the AppSyncClient configuration code to enable or suppress auto-reconnect behavior. If auto-reconnect is enabled, the AppSync client will reconnect subscriptions using an exponential backoff strategy (similar to what happens already in DeltaSync). If auto-reconnect is disabled, then a disconnect error will be propagated to the app via the subscription result handler, just as it is now, to be handled by the app.

palpatim:
(D'oh! Forgot to update status)

palpatim:
Pushed in #197.

This fixes the connection thrashing that can occur when the MQTT client
reconnects while AppSync is simultaneously advising clients to reconnect. The
preferred level at which to manage reconnections for AppSync is the AppSync
client itself, rather than the MQTTClient.

It's possible at some point in the future we may elect to expose an AppSync
client to support auto-reconnect, but even that would still be managed in the
AppSync client, rather than the MQTT client. It's also not a given that we want to expose that feature; this is something that the `sync` method manages on behalf of clients, and we'd prefer to steer clients toward using that method as much as possible rather.

muellerfr:
Closing issue after discussion with @palpatim.

PR #197, opened 2019-03-08
Palpatim/gh193 subscription reconnect

palpatim:
*Issue #, if available:*
#193

*Description of changes:*

**NOTE: This PR is based on #196. See https://github.com/awslabs/aws-mobile-appsync-sdk-ios/commit/42134bd6eac87b219e0314d36fda0c80e2da7e9c for the diff that applies just to this PR.**

This fixes the connection thrashing that can occur when the MQTT client
reconnects while AppSync is simultaneously advising clients to reconnect. The
preferred level at which to manage reconnections for AppSync is the AppSync
client itself, rather than the MQTTClient.

It's possible at some point in the future we may elect to expose an AppSync
client to support auto-reconnect, but even that would still be managed in the
AppSync client, rather than the MQTT client.


AWSAppSyncClient/MQTTSDK/AWSIoTMQTTClient.m
```diff
@@ -428,8 +432,12 @@ - (BOOL) webSocketConnectWithClientId {
         //Get Credentials from credentials provider.
         [[self.configuration.credentialsProvider credentials] continueWithBlock:^id _Nullable(AWSTask<AWSCredentials *> * _Nonnull task) {
             
-            //If an error occured when trying to get credentials, setup a timer to retry the connection after self.currentRecconectTime seconds and schedule it on the current Thread.
+            //If an error occured when trying to get credentials, setup a timer to retry the connection after self.currentReconnectTime seconds and schedule it on the current Thread.
             if (task.error) {
+                if (!SUPPORT_AUTO_RECONNECT) {
```
cbommas: set self.mqttStatus to AWSIoTMQTTStatusConnectionError and notify callback here

cbommas: Addressed

PR #210, opened 2019-03-29
Update development dependencies to AWS Core 2.9.3.

rohandubal:
*Description of changes:*

- Update development dependency to latest AWSCore version to be inline with official podspec.
- This change also pins `Reachability.swift`'s dependency to `4.3.0` since `4.3.1` sets the project dependency to Swift 5.0 which would require devs to have latest version of Xcode for development.


rohandubal:
closing in favor of #223

Issue #225, opened 2019-04-10
AWSAppSyncOfflineMutationDelegate sometimes isn't called


tony7126:
**Describe the bug**
Sometimes the AWSAppSyncOfflineMutationDelegate that we've set in our project isn't called.  I've narrowed it down (I think) to a race condition.   This guard statement is hit:
https://github.com/awslabs/aws-mobile-appsync-sdk-ios/blob/5025f67f2e574d8d17b71a51b6ea85841b8b1d55/AWSAppSyncClient/Internal/AWSPerformOfflineMutationOperation.swift#L74

Because self becomes nil before this async operation can take place.  This is happening because the OperationQueue which is retaining the Operation object releases objects when the object indicates it is in a ".finished" state.  This is called here:
https://github.com/awslabs/aws-mobile-appsync-sdk-ios/blob/5025f67f2e574d8d17b71a51b6ea85841b8b1d55/AWSAppSyncClient/Internal/AWSPerformOfflineMutationOperation.swift#L99

Which may or may not be called before the async operation is called:
https://github.com/awslabs/aws-mobile-appsync-sdk-ios/blob/5025f67f2e574d8d17b71a51b6ea85841b8b1d55/AWSAppSyncClient/Internal/AWSPerformOfflineMutationOperation.swift#L68

**To Reproduce**
1. Go to https://github.com/awslabs/aws-mobile-appsync-sdk-ios/blob/5025f67f2e574d8d17b71a51b6ea85841b8b1d55/AWSAppSyncClient/Internal/AWSPerformOfflineMutationOperation.swift#L68
2. Add a small sleep(X) operation right before it

**Expected behavior**
The delegate the user sets will never be called because self becomes nil

**Environment(please complete the following information):**
 - AppSync SDK Version: 2.10.3
 - Dependency Manager: Cocoapods
 - Swift Version : 4.2

**Device Information (please complete the following information):**
 - Device: iPhone6S, iPhone8
 - iOS Version: 12.2

PR #226, opened 2019-04-10
add to retain self

tony7126:
*Issue #225 :*

*Description of changes:*
Add withExtendedLifetime to retain self until potential delegate operation has been executed

### DIFF
--- a/AWSAppSyncClient/Internal/CachedMutationOperation.swift
+++ b/AWSAppSyncClient/Internal/CachedMutationOperation.swift
@@ -100,22 +100,23 @@ final class CachedMutationOperation: AsynchronousOperation, Cancellable {
 
     private func notifyCompletion(_ result: JSONObject?, error: Error?) {
         operationCompletionBlock?(self, error)
+        withExtendedLifetime(self) {
+            handlerQueue.async { [weak self] in
+                guard
+                    let self = self,
+                    let appSyncClient = self.appSyncClient,
+                    let offlineMutationDelegate = appSyncClient.offlineMutationDelegate
+                    else {
+                        return
+                }
 
-        handlerQueue.async { [weak self] in
-            guard
-                let self = self,
-                let appSyncClient = self.appSyncClient,
-                let offlineMutationDelegate = appSyncClient.offlineMutationDelegate
-                else {
-                    return
+                // call master delegate
+                offlineMutationDelegate.mutationCallback(
+                    recordIdentifier: self.mutation.recordIdentifier,
+                    operationString: self.mutation.operationString!,
+                    snapshot: result,
+                    error: error)
             }
-
-            // call master delegate
-            offlineMutationDelegate.mutationCallback(
-                recordIdentifier: self.mutation.recordIdentifier,
-                operationString: self.mutation.operationString!,
-                snapshot: result,
-                error: error)
         }
     }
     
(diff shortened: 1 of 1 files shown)


AWSAppSyncClient/Internal/AWSPerformOfflineMutationOperation.swift
```diff
@@ -64,22 +64,23 @@ final class AWSPerformOfflineMutationOperation: AsynchronousOperation, Cancellab
 
     private func notifyCompletion(_ result: JSONObject?, error: Error?) {
         operationCompletionBlock?(self, error)
+        withExtendedLifetime(self) {
+            handlerQueue.async { [weak self] in
```
sunchunqiang: Can you explain how the change works? 
withExtendedLifetime will make sure that self will not destroyed before its closure return, but here handlerQueue.async  will return before handlerQueue.async's closure return, so withExtendedLifetime still cannot make sure the closure of handlerQueue.async  is finished before self is destroyed

tony7126: I was under the impression that the scope (self) would be maintained as long as anything was unfinished with "withExtendedLifetime".  Is that wrong?  (If so can you link me to an explanation of this, if possible?).

Maybe a simpler option would be to add a third argument to notifyCompletion, a callback function.  Then the state of the object can be switched to finished in there.  I can make the update and push today if you think that would make more sense

tony7126: Ok I did a toy example just to confirm you were right (which you are):
https://gist.github.com/tony7126/16de12ea7360ebda9344d3ec413d9ba5

tony7126: Here is a working toy example of the callback solution:
https://gist.github.com/tony7126/7160067f7d7bc48cedf68c4421744f1d

Let me know if you think something similar to that makes sense

palpatim:
Assigning this to @rohandubal since this may conflict with recent mutation queue updates that they did.

rohandubal:
Hello @tony7126 

I had a look at the code changes and the gist you posted(nicely detailed and super helpful), I agree with code changes in the PR. What I still want to discuss is the primary objective of making this change. Why do you think it is beneficial to wait till the delegate notification is completely done before updating the status? I think that has a chance of opening up to certain race condition when `cancel` is invoked by the developer? What are your thoughts?

We would also need to rebase this branch if it has to be pulled in, but that can wait till we have the discussion about the purpose.

Best,
Rohan

tony7126:
Hey Rohan,
I assume you're referring to this check on the isCancelled flag (I only see 2 and the other looks pretty clearly safe)?
https://github.com/awslabs/aws-mobile-appsync-sdk-ios/blob/13c6a5026980a70dc50881e6f35a4314fffc0fc7/AWSAppSyncClient/Internal/AWSPerformOfflineMutationOperation.swift#L109

If the error is nil it should never hit that isCancelled check and there shouldn't be a race?  Or is there another possibility I'm missing?

rohandubal:
I agree with your changes and seems the right thing to do. Could you please rebase on top of current master?

tony7126:
rebased.  let me know if everything looks good.

tony7126:
@rohandubal 
any word here?

tony7126:
@rohandubal 
Just following up again.  Any reason this can't be merged in?

tony7126:
@rohandubal 
Not sure what should be done here.  Do you have any feedback?

tony7126:
Hi, I've looked at the latest master and it seems this logic still isn't in there.  Is there a reason not to add this in?  This addresses something I've definitely seen happen.

tony7126:
@rohandubal is there any reason to keep this open?

PR #233, opened 2019-04-16
Update Mutation behavior to be more resilient.

rohandubal:
*Description of changes:*

Updated logic for mutations includes:
- Remove dependency on reachability client
- Mutation request is always attempted now
- If the error is of domain NSURLErrorDomain and part of some error
codes, we do an exponential retry
- We interrupt the exponential retry if we get a network up event
- Mutation has a custom retry layer which specifically looks for
internet availability issues and retries.


AWSAppSyncClient/Internal/AWSPerformMutationOperation.swift
```diff
@@ -6,12 +6,19 @@
 
 import Foundation
 
+/// This class(operation) is responsible for executing mutations when they are submitted to the operation queue
+/// from application side and not when loading from cache. If the app is killed and restarted, these operations
+/// are lost. To ensure they still are sent, they are loaded from the cache and a different type of operation is
+/// created: `AWSPerformOfflineMutationOperation`.
 final class AWSPerformMutationOperation<Mutation: GraphQLMutation>: AsynchronousOperation, Cancellable {
```
palpatim: While we're in here, what about renaming these mutation classes to something more descriptive like `CachedMutationOperation` and `SessionMutationOperation` or similar? Basically, something to clearly differentiate between mutations that were invoked during the current app session and mutations loaded from cache, which therefore lose callback context?

AWSAppSyncClient/Internal/AWSPerformMutationOperation.swift
```diff
@@ -6,12 +6,19 @@
 
 import Foundation
 
+/// This class(operation) is responsible for executing mutations when they are submitted to the operation queue
```
palpatim: Suggested reword:

> A AWSPerformMutationOperation *(but see rename suggestion below)* is responsible for executing mutations when they are submitted from the application during the current session. If the app is killed and restarted, these operations are reloaded from cache, but their callback context will be lost, and instead the global delegate callback will be invoked with the mutation result. See `AWSPerformOfflineMutationOperation `.

AWSAppSyncClient/Internal/AWSPerformMutationOperation.swift
```diff
@@ -29,35 +36,61 @@ final class AWSPerformMutationOperation<Mutation: GraphQLMutation>: Asynchronous
...
         self.mutationResultHandler = mutationResultHandler
+        super.init()
+        determineMutationStep()
     }
 
     deinit {
         AppSyncLog.verbose("\(identifier ?? "(no identifier)"): deinit")
     }
+    
+    private func determineMutationStep() {
```
palpatim: If we do rename "next step" to "state", rename this to `resolveInitialMutationState` or similar

AWSAppSyncClient/Internal/AWSPerformMutationOperation.swift
```diff
@@ -69,35 +102,54 @@ final class AWSPerformMutationOperation<Mutation: GraphQLMutation>: Asynchronous
...
-    // MARK: Operation
-
-    override func start() {
-        if isCancelled {
-            state = .finished
-            return
+    
+    private func scheduleRetry() {
+        AppSyncLog.debug("Scheduling mutation retry for in-memory mutation.")
+        mutationRetryHelper = AWSMutationRetryHelper(retryAttemptNumber: currentAttemptNumber) {
```
palpatim: I believe this is going to result in a data race where the signal callback is being invoked on a different queue than `self` is being assigned.

AWSAppSyncClient/Internal/AWSPerformOfflineMutationOperation.swift
```diff
@@ -6,11 +6,18 @@
 
 import Foundation
 
+/// This class(operation) is responsible to handle mutations which were created from previous run of the app but
```
palpatim: Similar reword suggestion as above

AWSAppSyncClient/Internal/AWSPerformOfflineMutationOperation.swift
```diff
@@ -6,11 +6,18 @@
...
+/// were not able to sent to the service due to connectivity issues. These operations are unique and different
+/// from `AWSPerformMutationOperation` since they do not have an inmemory callback or an operation object they
+/// can be mapped to. They use json/ dictionary mechanism to pass data between service and developer callback.
 final class AWSPerformOfflineMutationOperation: AsynchronousOperation, Cancellable {
```
palpatim: Similar rename suggestion as above

AWSAppSyncClient/Internal/AWSPerformOfflineMutationOperation.swift
```diff
@@ -48,18 +66,39 @@ final class AWSPerformOfflineMutationOperation: AsynchronousOperation, Cancellab
...
+            // If the mutation next step is graphql operation, then we invoke the network send.
             send(mutation, completion: completion)
+        default:
+            break
+        }
+    }
+    
+     func scheduleRetry() {
+        AppSyncLog.debug("Scheduling mutation retry for persistent mutation.")
+        mutationRetryHelper = AWSMutationRetryHelper(retryAttemptNumber: currentAttemptNumber) {
```
palpatim: As above, I believe this is going to result in a data race

AWSAppSyncTestApp/AppDelegate.swift
```diff
@@ -2,12 +2,14 @@
...
-//  Created by Dubal, Rohan on 9/11/18.
-//  Copyright © 2018 Amazon Web Services. All rights reserved.
+// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// Licensed under the Amazon Software License
+// http://aws.amazon.com/asl/
 //
 
 import UIKit
 
+
```
palpatim: Nit, remove extra line

Podfile
```diff
@@ -6,7 +6,7 @@ use_frameworks!
 target 'AWSAppSync' do
   pod 'AWSCore', '~> 2.9.0'
   pod 'SQLite.swift', '0.11.5'
-  pod 'ReachabilitySwift', '~> 4.3.0'
+  pod 'ReachabilitySwift', '4.3.0'
```
palpatim: Please add a comment explaining why we pinned this version

AWSAppSyncClient/AWSAppSyncClientS3ObjectsExtensions.swift
```diff
@@ -7,39 +7,14 @@ import Foundation
...
-                    conflictResolutionBlock: conflictResolutionBlock,
-                    handlerQueue: handlerQueue,
-                    resultHandler: resultHandler)
-            } else {
-                resultHandler?(nil, error)
-            }
+        resultHandler: @escaping (Error?) -> Void) {
+        
+        s3ObjectManager!.upload(s3Object: s3Object) { _, error in
+            resultHandler(error)
```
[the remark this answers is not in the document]
rohandubal: The expectation is that one of the two is set in the completion handler.

AWSAppSyncClient/Internal/AWSMutationRetryAdviceHelper.swift
```diff
@@ -0,0 +1,45 @@
...
+
+final class AWSMutationRetryAdviceHelper {
+    
+    static func isErrorRetriable(error: Error) -> Bool {
+        if let appsyncError = error as? AWSAppSyncClientError {
+            switch appsyncError {
+            case .authenticationError(let authError):
+                return isErrorURLDomainError(error: authError)
+            case .requestFailed(_, _, let urlError):
+                if urlError != nil {
```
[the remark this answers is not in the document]
rohandubal: yup will fix

AWSAppSyncClient/Internal/AWSMutationRetryAdviceHelper.swift
```diff
@@ -0,0 +1,45 @@
...
+
+import Foundation
+
+final class AWSMutationRetryAdviceHelper {
+    
+    static func isErrorRetriable(error: Error) -> Bool {
+        if let appsyncError = error as? AWSAppSyncClientError {
+            switch appsyncError {
+            case .authenticationError(let authError):
+                return isErrorURLDomainError(error: authError)
```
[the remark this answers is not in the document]
rohandubal: Yes I caught this in IAM auth test (Cognito Identity SDK returned this error which we wrap it.)

Yes, going to document it and add a test.

AWSAppSyncClient/Internal/AWSMutationRetryAdviceHelper.swift
```diff
@@ -0,0 +1,45 @@
...
+// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// Licensed under the Amazon Software License
+// http://aws.amazon.com/asl/
+//
+
+import Foundation
+
+final class AWSMutationRetryAdviceHelper {
+    
+    static func isErrorRetriable(error: Error) -> Bool {
```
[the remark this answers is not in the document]
rohandubal: The renaming makes sense. Will update to `isRetriableNetworkError`

AWSAppSyncClient/Internal/AWSMutationRetryAdviceHelper.swift
```diff
@@ -0,0 +1,45 @@
...
+            }
+        } else {
+            return isErrorURLDomainError(error: error)
+        }
+        return false
+    }
+    
+    /// We evaluate the error against known error codes which could result due to unavailable internet or spotty network connection.
+    private static func isErrorURLDomainError(error: Error) -> Bool {
+        if let nserror = error as NSError?, nserror.domain == NSURLErrorDomain {
```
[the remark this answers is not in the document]
rohandubal: Will update. It is indeed due to the casting issue you pointed out.

AWSAppSyncClient/Internal/AWSMutationRetryHelper.swift
```diff
@@ -0,0 +1,93 @@
...
+
+/// This class is responsible to create a timer and watch for network up events and
+/// retry the mutation when the timer expires or network up event is received.
+/// Note: -  This class can also be used by `sync` to watch for network events and back-off based on timer in future.
+final class AWSMutationRetryHelper: Cancellable {
+    private var nextSyncTimer: DispatchSourceTimer?
+    private var retryAttemptNumber: Int
+    private var retryStrategy: AWSAppSyncRetryStrategy = .exponential
+    /// Ensures ordered callback of retry notification dispatch in case of multiple attempts.
+    private var retryOperationQueue: OperationQueue
```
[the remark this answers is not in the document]
rohandubal: Will add the label. Will evaluate to ensure there are no concerns in using serial DispatchQueue and make the change.

AWSAppSyncClient/Internal/AWSMutationRetryHelper.swift
```diff
@@ -0,0 +1,93 @@
...
+         retrySignalCallback: @escaping () -> Void) {
+        self.retryAttemptNumber = retryAttemptNumber
+        self.retrySignalCallback = retrySignalCallback
+        
+        retryOperationQueue = OperationQueue()
+        retryOperationQueue.maxConcurrentOperationCount = 1
+        
+        let delayForCurrentAttempt = AWSAppSyncRetryHandler.retryDelayInMillseconds(for: retryAttemptNumber, retryStrategy: retryStrategy)
+        // Either follow the advice from retry strategy or fallback to max wait.
+        let delay = min(delayForCurrentAttempt, AWSAppSyncRetryHandler.maxWaitMilliseconds)
```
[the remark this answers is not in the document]
rohandubal: Currently that logic is decoupled to break the retries for 4xx and 5xx errors. In the retry handler if the delay exceeds 5 minutes then advice becomes `false`.  Hence need to do it explicitly here to ensure we never exhaust retries.

AWSAppSyncClient/Internal/AWSMutationRetryHelper.swift
```diff
@@ -0,0 +1,93 @@
...
+    deinit {
+        cancel()
+        // Defensively remove from any remaining notifications, but this should have already been handled in
+        // `unregisterForNotifications()`
+        NotificationCenter.default.removeObserver(self)
+    }
+    
+    private func scheduleTimer(at deadline: DispatchTime) {
+        nextSyncTimer = DispatchSource.makeOneOffDispatchSourceTimer(deadline: deadline, queue: handlerQueue) {
+            AppSyncLog.debug("Timer fired, attempting mutation operation")
```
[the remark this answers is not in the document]
rohandubal: Will add.

AWSAppSyncClient/Internal/AWSMutationRetryHelper.swift
```diff
@@ -0,0 +1,93 @@
...
+    }
+    
+    func notifyCallback() {
+        // Call the cancel routine as the purpose of retry is fulfilled
+        cancel()
+        retrySignalCallback()
+    }
+    
+    @objc private func didConnectivityChange(notification: Notification) {
+        // If internet was disconnected and is available now, perform deltaSync
```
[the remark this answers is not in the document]
rohandubal: haha totally missed this. will update.

AWSAppSyncClient/Internal/AWSMutationRetryHelper.swift
```diff
@@ -0,0 +1,93 @@
...
+// Licensed under the Amazon Software License
+// http://aws.amazon.com/asl/
+//
+
+import Foundation
+
+/// This class is responsible to create a timer and watch for network up events and
+/// retry the mutation when the timer expires or network up event is received.
+/// Note: -  This class can also be used by `sync` to watch for network events and back-off based on timer in future.
+final class AWSMutationRetryHelper: Cancellable {
```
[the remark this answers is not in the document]
rohandubal: Re. 1 - My intention was to keep it other way round. The use case of this class would be to notify the operation/ caller who is waiting for a specific set of conditions to occur and get notified when it happens. That way we have a central place for our different use-cases where network notifications and timer events are handled.
2. Agreed, `AWSMutationRetryNotifier` makes it more clear.

AWSAppSyncClient/Internal/AWSMutationSteps.swift
```diff
@@ -0,0 +1,16 @@
...
+//
+
+import Foundation
+
+/// Determines the next step in a mutation operation.
+///
+/// - unknown: when the next steo of mutation is not determined yet
+/// - s3Upload: the mutation is required to do a s3 upload before the graphql call
+/// - graphqlOperation: the mutation to complete needs to make a graphql call
+enum MutationNextStep {
```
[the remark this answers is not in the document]
rohandubal: Will refactor.

AWSAppSyncClient/Internal/AWSPerformMutationOperation.swift
rohandubal: I like the idea. I was slightly confused and thus the documentation. But renaming should alleviate some issues.

Podfile
rohandubal: Will add.

AWSAppSyncClient/Internal/AWSPerformMutationOperation.swift
rohandubal: I will investigate this and follow up.

rohandubal: I looked and I think I understand how it could result in data race(less likely currently due to the wait from the timer but still CAN happen). I will update the code to ensure that we do not run into that scenario. Great catch!

AWSAppSyncClient/Internal/AWSMutationRetryNotifier.swift
```diff
@@ -54,7 +58,7 @@ final class AWSMutationRetryHelper: Cancellable {
     
     private func scheduleTimer(at deadline: DispatchTime) {
         nextSyncTimer = DispatchSource.makeOneOffDispatchSourceTimer(deadline: deadline, queue: handlerQueue) {
-            AppSyncLog.debug("Timer fired, attempting mutation operation")
+            AppSyncLog.debug("Timer fired, attempting mutation operation. Retry number: \(self.retryAttemptNumber)")
```
palpatim: nit, store `retryAttemptNumber` outside closure to avoid referencing `self` within it.

AWSAppSyncClient/Internal/AWSMutationRetryAdviceHelper.swift
```diff
@@ -8,14 +8,20 @@ import Foundation
 
 final class AWSMutationRetryAdviceHelper {
     
-    static func isErrorRetriable(error: Error) -> Bool {
+    static func isRetriableNetworkError(error: Error) -> Bool {
```
palpatim: Just to confirm; we will also retry for service level errors, but that retry logic is handled at the network transport layer, right?

I'd like to centralize the retry handling, but I'd guess that's not easily accomplishable, so let's at least document the different retry handling levels some place so it's easier to reason about reviewing the mutation operation code. Perhaps this could be done in the `send` method of the mutation operations, as the top level entry point into the various layers of retry.

AWSAppSyncClient/Internal/CachedMutationOperation.swift
```diff
@@ -6,20 +6,17 @@
...
-/// from `AWSPerformMutationOperation` since they do not have an inmemory callback or an operation object they
-/// can be mapped to. They use json/ dictionary mechanism to pass data between service and developer callback.
-final class AWSPerformOfflineMutationOperation: AsynchronousOperation, Cancellable {
+/// This class(operation) is responsible for executing mutations when they are submitted from the application during the previous session, but were not be able to sent to service due to network not being available or app being quit. The callbacks for these mutations are given to the global callback delegate. For mutations submitted in current session, see: `SessionMutationOperation`.
+final class CachedMutationOperation: AsynchronousOperation, Cancellable {
     private weak var appSyncClient: AWSAppSyncClient?
     private weak var networkClient: AWSNetworkTransport?
     private let handlerQueue: DispatchQueue
     let mutation: AWSAppSyncMutationRecord
     var currentAttemptNumber: Int = 1
```
palpatim: nit: don't need to specify `Int` :)

AWSAppSyncClient/Internal/SessionMutationOperation.swift
```diff
@@ -6,24 +6,21 @@
...
-/// created: `AWSPerformOfflineMutationOperation`.
-final class AWSPerformMutationOperation<Mutation: GraphQLMutation>: AsynchronousOperation, Cancellable {
+/// This class(operation) is responsible for executing mutations when they are submitted from the application during the current session. If the app is killed and restarted, these operations are reloaded from cache, but their callback context will be lost, and instead the global delegate callback will be invoked with the mutation result. See: `CachedMutationOperation`.
+final class SessionMutationOperation<Mutation: GraphQLMutation>: AsynchronousOperation, Cancellable {
     private weak var appSyncClient: AWSAppSyncClient?
     private let handlerQueue: DispatchQueue
     private let mutation: Mutation
     private let mutationConflictHandler: MutationConflictHandler<Mutation>?
     private let mutationResultHandler: OperationResultHandler<Mutation>?
     var currentAttemptNumber: Int = 1
```
palpatim: As above, no need to specify `Int`

CHANGELOG.md
```diff
@@ -2,6 +2,23 @@
...
+## Misc. Updates
+
+**Behavior Change for Mutation Queue**
+
+- With this update, the internal operation logic of how mutations are processed is updated
+- The mutation queue in the previous SDKs, depended on notification from reachability to determine if the mutations could be sent to the service; in the new behavior we are more aligned to Apple's [Designing for Real-World Networks
+](https://developer.apple.com/library/archive/documentation/NetworkingInternetWeb/Conceptual/NetworkingOverview/WhyNetworkingIsHard/WhyNetworkingIsHard.html) guidance
+- The AppSync client will now `always` attempt to make a mutation request regardless of the network state
+- One the network response comes back, the SDK will inspect the error from `NSURLSession` and determine if the error was due to network not available, host not found or DNS lookup failed. If it was, the SDK will schedule a retry timer responsible to retry the request which will grow exponentially with every attempt
+- The retry handler will also watch for notification from reachability to determine if network is available again; in cases where SDK does get the notification, it will prempt the timer and make the request right away
```
palpatim: s/prempt/preempt/

CHANGELOG.md
```diff
@@ -2,6 +2,23 @@
...
+- If the notification is not received, the timer will contrinue to retry the request. The polling time is capped at 5 minutes to ensure that attempts are made at frequent attempts while respecting resources on device
```
palpatim: 1. s/contrinue/continue/
1. Please clarify that the polling *interval* is capped at 5 minutes, and that retries will continue every 5 minutes until cancelled or successful

rohandubal: Good point. Will update the wording.

CHANGELOG.md
rohandubal: argh these typos

AWSAppSyncClient/Internal/SessionMutationOperation.swift
rohandubal: Will update

AWSAppSyncClient/Internal/AWSMutationRetryAdviceHelper.swift
rohandubal: Yes, you are right. I will update the documentation.

PR #245, opened 2019-04-29
Make network reachability provider mockable

gleue:
*Description of changes:*

Merging this PR will make

* **initializer** `AWSAppSyncClient.init(appSyncConfig:reachabilityFactory:)` public, and
* **protocols** `NetworkReachabilityProvidingFactory` and `NetworkReachabilityProviding` public

which are all `internal` to the framework for now.

*Rationale behind changes:*

By allowing framework users access to above API it will be possible to

* **mock** the reachability service properly for integration tests
* **substitute** the reachability service with a private solution if already present in a project

*Test results*

The proposed changes have no effect on unit or integration tests. All tests succeed w/ changes
applied.

### DIFF
--- a/AWSAppSyncClient/AWSAppSyncClient.swift
+++ b/AWSAppSyncClient/AWSAppSyncClient.swift
@@ -66,8 +66,8 @@ public class AWSAppSyncClient {
         try self.init(appSyncConfig: appSyncConfig, reachabilityFactory: nil)
     }
 
-    init(appSyncConfig: AWSAppSyncClientConfiguration,
-         reachabilityFactory: NetworkReachabilityProvidingFactory.Type? = nil) throws {
+    public init(appSyncConfig: AWSAppSyncClientConfiguration,
+                reachabilityFactory: NetworkReachabilityProvidingFactory.Type? = nil) throws {
 
         AppSyncLog.info("Initializing AppSyncClient")
 
--- a/AWSAppSyncClient/AWSAppSyncClient.swift
+++ b/AWSAppSyncClient/AWSAppSyncClient.swift
@@ -66,6 +66,19 @@ public class AWSAppSyncClient {
         try self.init(appSyncConfig: appSyncConfig, reachabilityFactory: nil)
     }
 
+    /// Creates a client with the specified `AWSAppSyncClientConfiguration`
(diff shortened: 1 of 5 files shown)


AWSAppSyncClient/Internal/NetworkReachability.swift
```diff
@@ -8,14 +8,14 @@ import Foundation
 import Reachability
 
 /// Defines a factory to return a NetworkReachabilityProviding instance
-protocol NetworkReachabilityProvidingFactory {
+public protocol NetworkReachabilityProvidingFactory {
```
palpatim: Please move the newly-public protocols and extensions out of the `Internal` folder into a top-level file called `NetworkReachability.swift`, and let's rename this file to `Internal/NetworkReachabilityNotifier.swift`

AWSAppSyncClient/AWSAppSyncClient.swift
```diff
@@ -66,8 +66,8 @@ public class AWSAppSyncClient {
         try self.init(appSyncConfig: appSyncConfig, reachabilityFactory: nil)
     }
 
-    init(appSyncConfig: AWSAppSyncClientConfiguration,
-         reachabilityFactory: NetworkReachabilityProvidingFactory.Type? = nil) throws {
+    public init(appSyncConfig: AWSAppSyncClientConfiguration,
```
palpatim: I agree in principle with the change, but in order to make the intent clear, let's add some documentation along the lines of:

```swift
/// Creates an AWSAppSyncClient with the specified configuration and NetworkReachabilityProvidingFactory.
/// This method is primarily intended to facilitate integration testing, but may be used by apps that wish
/// to use their own reachability solution in place of AppSync's bundled reachability framework.
///
/// Note that the AppSync client's use of the Reachability client provided by the factory should be
/// considered an implementation detail. In particular, apps should not rely on AppSync to inspect network
/// reachability status before attempting a network connection.
///
/// - Parameters:
///   - appSyncConfig: The `AWSAppSyncClientConfiguration` object.
///   - reachabilityFactory: The factory that provides `NetworkReachabilityProviding` instances
```

palpatim:
Thanks for the PR!

Issue #249, opened 2019-05-02
Error in restarting subscription after cancellation


Laxmikantwagh:
**Describe the bug**
When app goes to background , we cancel the subscription. Next time, when app moves to foreground, we restart subscription. But sometimes, when app comes to foreground/ active state, the request of restarting subscription does not work. At this time, there is no error callbacks. 

**To Reproduce**
1. Launch the app (subscription is activated as app is in active state).
2. Now, lock the device and then unlock it.
3. Step 2 should be repeated 3-4 times quickly.
4. Unlock the device.
5. Now, the subscription never starts.

**Expected behavior**
If app is moved to background and foreground multiple times and then if app is in foreground/active state finally, the request to restart subscription should be successful. 

**Environment**
 - AppSync SDK Version: 2.10.3
 - Dependency Manager: Cocoapods
 - Swift Version : 4.2

**Device Information:**
 - Device: iPhone 6, iPhone X
 - iOS Version: 12.2

**More Information**
We have just tried one thing.
class -> **AWSAppSyncSubscriptionWatcher.swift**
method -> init

In this method, self.startSubscription() was not being called sometimes. So, we have updated this method. 

PFA screen shot for more details.

<img width="570" alt="Screenshot 2019-05-02 at 3 47 35 PM" src="https://user-images.githubusercontent.com/45625291/57069345-fc75af80-6cf1-11e9-90d8-f93717a9dcca.png">

**Conclusion:**
Please suggest on this bug. 
Also, please check our approach. Do you have any better way to fix this issue ?

shannon-hager-skookum:
@Laxmikantwagh The AppSync SDK version listed above is 2.10.3, I would update to the current version (2.12.1) and see if the problem is still present. There has been some work on this area of functionality that would may solve your issue and if not, might affect your workaround.

palpatim:
This is released on https://github.com/awslabs/aws-mobile-appsync-sdk-ios/releases/tag/2.12.2

Please let us know if you have any further issues.

Laxmikantwagh:
Thank you so much 👍 . I will check it and get back if required.

PR #251, opened 2019-05-08
Fix a bug which blocks subscription from starting after being cancelled.

rohandubal:
*Issue #, if available:*
#249 

*Description of changes:*
- Signal the semaphore from `cancel` method since we do not want to block any other subscriptions
- Added unit test which validates the behavior
- Added functional test which validates the behavior with actual subscription requests

### DIFF
--- a/AWSAppSyncClient/AWSAppSyncSubscriptionWatcher.swift
+++ b/AWSAppSyncClient/AWSAppSyncSubscriptionWatcher.swift
@@ -52,7 +52,6 @@ private class SubscriptionsOrderHelper {
     
 }
 
-
 /// Used to determine the reason why a subscription is being cancelled/ disconnected.
 ///
 /// - none: Indicates there is no source of cancellation yet.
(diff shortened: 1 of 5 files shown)


AWSAppSyncClient/AWSAppSyncSubscriptionWatcher.swift
```diff
@@ -220,6 +222,10 @@ public final class AWSAppSyncSubscriptionWatcher<Subscription: GraphQLSubscripti
     internal func performCleanUpTasksOnCancel() {
         isCancelled = true
         status = .disconnected
+        // We signal the semaphore here to ensure that if the subscription request is waiting for HTTP call to return,
+        // we do not block on it. The subscription is already cancelled and we can release the wait for other
+        // subscriptions to resume.
+        self.semaphore.signal()
```
palpatim: nit: remove `self`

rohandubal: Done.

PR #255, opened 2019-05-14
Add multiclient support

minbi:


AWSAppSyncClient/AWSAppSyncClient.swift
```diff
@@ -126,7 +126,7 @@ public class AWSAppSyncClient {
                 let prefixTrackerValue = appSyncConfig.url.absoluteString + "_" + authTypeString
                 if let (clientString, clientCount) = AWSAppSyncClient.prefixTracker[prefixTrackerKey] {
                     if clientString != prefixTrackerValue {
-                        throw AWSAppSyncClientConfigurationError.invalidClientDatabasePrefixConfiguration("Configured two clients with the same database prefix")
+                        throw AWSAppSyncClientConfigurationError.clientDatabasePrefixAlreadyInUse("Configured two clients with the same database prefix")
```
rohandubal: recommend: `AWSAppSyncClientConfigurationError.cacheConfiguraitonAlreadyInUse`
This will reduce the need to add another enum for future validations where 2 clients have no prefix

AWSAppSyncClient/AWSAppSyncClient.swift
```diff
@@ -184,9 +183,12 @@ public class AWSAppSyncClient {
         } catch {
```
rohandubal: Need to validate if Promise resolved to failure

AWSAppSyncClient/AWSAppSyncClientConfigurationError.swift
```diff
@@ -22,7 +22,7 @@ extension AWSAppSyncClientConfigurationError: LocalizedError {
         switch self {
         case .invalidAuthConfiguration(let message):
             return "Invalid Auth Configuration: \(message)"
-        case .invalidClientDatabasePrefixConfiguration(let message):
+        case .clientDatabasePrefixAlreadyInUse(let message):
             return "Invalid Client Database Prefix Configuration: \(message)"
```
rohandubal: message should reflect that the cache configuration is in use.

AWSAppSyncClient/Internal/AWSPerformMutationQueue.swift
```diff
@@ -162,9 +162,9 @@ final class AWSPerformMutationQueue {
             .catch { error in AppSyncLog.error("\(#function) failure: \(error)") }
     }
 
-    func clearQueue() {
+    func clearQueue() throws {
         operationQueue.cancelAllOperations()
-        _ = persistentCache?.clear()
+        try persistentCache?.clear().await()
```
rohandubal: same as above. Need to ensure promise is not resolved to error case.

AWSAppSyncUnitTests/AWSAppSyncCacheConfigurationTests.swift
```diff
@@ -114,16 +114,17 @@ class AWSAppSyncCacheConfigurationTests: XCTestCase {
         )
         let cacheConfig = try AWSAppSyncCacheConfiguration(useClientDatabasePrefix: true, appSyncServiceConfig: serviceConfigWithFooPrefix)
         let clientConfig = try AWSAppSyncClientConfiguration(appSyncServiceConfig: serviceConfigWithFooPrefix, cacheConfiguration: cacheConfig)
-        let client = try AWSAppSyncClient(appSyncConfig: clientConfig)
-        let clientb = try AWSAppSyncClient(appSyncConfig: clientConfig)
+        let clientA = try AWSAppSyncClient(appSyncConfig: clientConfig)
```
rohandubal: can we use `let _ = ....`?

minbi: can't this triggers dealloc code

rohandubal: That should not affect the `try` logic test. The `dealloc` would be done once the init for `AWSAppSyncClient` is complete I believe.

AWSAppSyncClient/AWSAppSyncClientConfigurationError.swift
minbi: changed.

AWSAppSyncClient/AWSAppSyncClient.swift
minbi: await should throw the error if failure

AWSAppSyncClient/Internal/AWSPerformMutationQueue.swift
minbi: await should throw in failure case

AWSAppSyncUnitTests/AWSAppSyncCacheConfigurationTests.swift
minbi: It doesn't affect clientA or clientB it affects the last client construction which expects a failure because the prefix is used. By using the `_`, the clients are de-allocated before the final client can test a failure case.

PR #256, opened 2019-05-14
Update CHANGELOG.md

minbi:


CHANGELOG.md
```diff
@@ -3,6 +3,40 @@
 The AWS AppSync SDK for iOS enables you to access your AWS AppSync backend and perform operations like `Queries`, `Mutations` and `Subscriptions`. The SDK
 also includes support for offline operations.
 
+## 2.13.0
```
rohandubal: Can you callout why this is a `y` version bump?

CHANGELOG.md
```diff
@@ -3,6 +3,40 @@
...
+          "Region": "us-east-2",
+          "AuthMode": "API_KEY",
+          "ApiKey": "da2-xyz",
+          "ClientDatabasePrefix": "MyAppSyncAPIName_API_KEY"
+        }
+      }
+      ```
+
+      The `AWSAppSyncClient` object can be constructed as follows:
+      ```swift
```
rohandubal: I think this code block can be confusing. There are places where we accept `nil` values and in this case we should document which values are expected to be non-nil.

PR #261, opened 2019-05-21
Update subscriptions order helper logic.

rohandubal:
*Issue*
- A customer reported that they are seeing a crash when accessing the dictionary in `SubscriptionsOrderHelper` when `4` subscription calls are made in quick succession
- Adding a delay of `0.1s` resolved the issue for them, but it is a temporary workaround
- The issue cannot be reproduced via functional tests
- The issue happens randomly in app when delay is not added, can be reproduced but not consistently

*Description of changes:*
- Improve the logic of how `SubscriptionsOrderHelper` vends unique IDs to watchers
- All existing tests pass
- There were some unused methods in `SubscriptionsOrderHelper` which are now deleted

### DIFF
--- a/AWSAppSyncClient/AWSAppSyncSubscriptionWatcher.swift
+++ b/AWSAppSyncClient/AWSAppSyncSubscriptionWatcher.swift
@@ -25,13 +25,13 @@ import os.log
     func subscriptionAcknowledgementDelegate()
 }
 
-private class SubscriptionsOrderHelper {
-    private var count = 0
-    let serialQueue = DispatchQueue(label: "com.amazonaws.appsync.subscriptionsOrderHelper")
+internal class SubscriptionsOrderHelper {
+    private static var count = 0
+    static let serialQueue = DispatchQueue(label: "com.amazonaws.appsync.subscriptionsOrderHelper")
     
-    static let sharedInstance = SubscriptionsOrderHelper()
+    private init() {}
     
-    var uniqueIdentifier: Int {
+    static func getNextUniqueIdentifier() -> Int {
         return serialQueue.sync {
             count += 1
             return count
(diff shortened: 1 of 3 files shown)


AWSAppSyncClient/AWSAppSyncSubscriptionWatcher.swift
```diff
@@ -26,30 +26,17 @@ import os.log
...
     
-    func markDone(id: Int) {
-        waitDictionary[id] = true
-    }
+    static let sharedInstance = SubscriptionsOrderHelper()
     
-    func shouldWait(id: Int) -> Bool {
-        for i in 0..<id where waitDictionary[i] == false {
-            return true
+    var uniqueIdentifier: Int {
```
palpatim: 1. Thanks for cleaning up the unused properties
1. Accessing properties shouldn't mutate. Make this a method like `getNextUniqueIdentifier()`
1. It feels a bit heavyweight to use a `DispatchQueue` for synchronizing a counter that you never actually read other than when obtaining the next value. Would a `DispatchSempahore` suffice here?
1. Why not make the counter itself a static member; the instance has no particular use here since it's never overwritten once instantiated
    1. Sub-note: I don't recall that static integers are inherently atomic, so it's probably still necessary to do multi-threading protection, but if you're aware that statics are in fact atomic, then this whole thing pretty much devolves into a static `Int` counter on the `AWSAppSyncSubscriptionWatcher` class.

rohandubal: - You are right about the static int not being atomic and thus this approach.
- I agree with having a method and will add one
- I can use a dispatch semaphore, but I chose the DispatchQueue based on my initial reading and seeing what the community is doing. Will dive deep a little more to be sure if I can use DispatchSemaphore
- I'll also consider making the method static

rohandubal: `DispatchSemaphore` cannot be used since it would block the calling thread (which could be different in our case)

I am not looking at `DispatchGroup` to validate if it can be used. If not, I might have to fallback on using a queue.

rohandubal: Added a test as well to test the behavior.

PR #280, opened 2019-07-23
 Bump AWSAppSync.podspec dependency on AWSCore

phanisrikar93:
Bump AWSAppSync dependency on AWSCore to 2.10.0.
This change would enable developers to use the latest version of AWSAppSync with the latest versions of other Mobile SDKs.

### DIFF
--- a/AWSAppSync.podspec
+++ b/AWSAppSync.podspec
@@ -11,7 +11,7 @@ Pod::Spec.new do |s|
   s.ios.deployment_target = '9.0'
   s.swift_version = '4.2'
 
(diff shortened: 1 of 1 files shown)


palpatim:
Thanks for the PR. Unfortunately it didn't go quite far enough--needed to update the Podspec, Podfile, Podfile.lock, and committed files as well. I've done that for 2.10.2 in https://github.com/awslabs/aws-mobile-appsync-sdk-ios/pull/285, so I'm closing this PR in favor of that one.

Issue #287, opened 2019-08-06
Crashes in [AWSMQTTSession decoder:newMessage:]


adamup928:
**Describe the bug**
We're seeing occasional crashes when new messages are received by AWSMQTTSession.  They are infrequent and we are not able to reproduce.  This is occurring while AppSync is in active use and the app is in the active, foreground state (not dismissing to or returning from background).

**Environment(please complete the following information):**
 - AppSync SDK Version: 2.10.4
 - Dependency Manager: Cocoapods
 - Swift Version: 4.0

**Additional context**

Crash log:
```
# Cause: Crash due to signal: SIGSEGV(SEGV_MAPERR) at 00000008
```

```
Thread 28 [Crashed]:

0    CoreFoundation                           0x1c4bd77d0     CFRunLoopAddTimer + 136
1    AWSAppSync                               0x102df5e20     -[AWSMQTTSession decoder:newMessage:] (AWSMQTTSession.m:389)
2    AWSAppSync                               0x102df1f1c     -[AWSMQTTDecoder stream:handleEvent:] (AWSMQTTDecoder.m:133)
3    CoreFoundation                           0x1c4bedbc8     _signalEventSync + 216
4    CoreFoundation                           0x1c4bed1bc     _cfstream_shared_signalEventSync + 372
5    CoreFoundation                           0x1c4bdb018     _CFRUNLOOP_IS_CALLING_OUT_TO_A_SOURCE0_PERFORM_FUNCTION__ + 24
6    CoreFoundation                           0x1c4bdaf98     _CFRunLoopDoSource0 + 88
7    CoreFoundation                           0x1c4bda880     _CFRunLoopDoSources0 + 176
8    CoreFoundation                           0x1c4bd57bc     _CFRunLoopRun + 1004
9    CoreFoundation                           0x1c4bd50b0     CFRunLoopRunSpecific + 436
10   Foundation                               0x1c55a2fac     -[NSRunLoop(NSRunLoop) runMode:beforeDate:] + 300
11   AWSAppSync                               0x102decc4c     -[AWSIoTMQTTClient openStreams:] (AWSIoTMQTTClient.m:660)
12   Foundation                               0x1c56cf6a4     _NSThread__start__ + 984
13   libsystem_pthread.dylib                  0x1c48682c0     _pthread_body + 128
14   libsystem_pthread.dylib                  0x1c4868220     _pthread_start + 44
[...]

PR #295, opened 2019-09-06
Palpatim/troubleshoot subscription crash

palpatim:
*Issue #, if available:*
#287 

*Description of changes:*

Attempting to troubleshoot crashes reported on #287, this PR does a few things:

- Updates AWS dependencies to the latest version
- Adds a "lifecycle" test to ensure that subscription callbacks are not invoked because of a customer-initiated cancellation request
- Removes some unused code from the IoT module
- Updates the test app to allow us to use it to test subscription connections/disconnections while mutations are occurring
- Renamed streams inside of MQTTClient/Session and clarified their usage in a README doc


palpatim:
Now that we have direct websocket connection and are not using MQTT/IoT, this PR is no longer valid.

PR #305, opened 2019-10-15
Fix race condition in testCancelAllMutations

palpatim:
Fixed race condition waiting for operation queue to be emptied

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/AWSAppSyncUnitTests/MutationQueueTests.swift
+++ b/AWSAppSyncUnitTests/MutationQueueTests.swift
@@ -641,34 +641,57 @@ class MutationQueueTests: XCTestCase {
         XCTAssertEqual(n0, n2)
     }
 
+    /// This test asserts two behaviors:
+    ///
+    /// Test 1
+    /// - Given: A queue of in-process mutations
+    /// - When: I invoke `clearCaches` with the option set to clear the mutation queue
+    /// - Then: The mutation queue is emptied
+    ///
+    /// Test 2
+    /// - Given: A queue of in-process mutations
+    /// - When: I clear the cache, then perform another mutation
+    /// - Then: The new mutation is performed as expected
     func testCancelAllMutations() throws {
+        // Test 1, which is also setup for Test 2
         let addPost = DefaultTestPostData.defaultCreatePostWithoutFileUsingParametersMutation
 
         let mockHTTPTransport = MockAWSNetworkTransport()
-        mockHTTPTransport.operationResponseDelay = 5
+        mockHTTPTransport.operationResponseDelay = 2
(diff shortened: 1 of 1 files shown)


AWSAppSyncUnitTests/MutationQueueTests.swift
```diff
@@ -641,34 +641,57 @@ class MutationQueueTests: XCTestCase {
...
+                queuedMutationShouldNotBeInvoked.fulfill()
             }
         }
 
         try appSyncClient.clearCaches(options: ClearCacheOptions(clearMutations: true))
 
-        wait(for: [mutationPerformed], timeout: 6.0)
+        wait(for: mutationExpectations, timeout: 3.0)
 
         XCTAssertEqual(appSyncClient.queuedMutationCount, 0)
```
lawmicha: would be nice to see use of assert overload with message like "The mutation queue should be emptied"

lawmicha: would be nice to see use of assert overload with message like "The mutation queue should be emptied"

AWSAppSyncUnitTests/MutationQueueTests.swift
```diff
@@ -641,34 +641,57 @@ class MutationQueueTests: XCTestCase {
...
 
         mockHTTPTransport.operationResponseDelay = 0
 
+        // Test 2
         let furtherMutationPerformed = expectation(description: "Further post added")
         appSyncClient.perform(mutation: addPost) { result, error in
             XCTAssertEqual(result?.data?.createPostWithoutFileUsingParameters?.id, "TestPostID")
```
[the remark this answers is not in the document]
palpatim: Good callout. I'll take that on as a future enhancement.

AWSAppSyncUnitTests/MutationQueueTests.swift
palpatim: See comment above--I think this is a dupe

lawmicha: yes for the `XCTAssertEqual`, okay makes sense! it was just initial intuition on my part.

just thinking out loud here, i wonder if this improves the test durability with an assert on the first mutation being executed.
```
// makes sure the first mutation was added is completed
wait(for: firstMutationExpectation, timeout: 1.0)

// check that the total mutations minus one is in the queue
XCTAssertEqual(appSyncClient.queuedMutationCount, expectedQueuedMutationCount)

// clear the cache
try appSyncClient.clearCaches(...)

// make sure the inverted expectations are not fulfilled after mutation queue is cleared
wait(for: mutationExpectations, timeout: 3.0)

// make sure there are no queued mutations
XCTAssertEqual(appSyncClient.queuedMutationCount, 0)
```

The reasoning is, seems unlikely, but there is a race condition on clearing the cache and asserting/fulfilling the first mutation.

PR #337, opened 2020-02-18
AppSyncClient and Amplify API subscription code consolidation

lawmicha:
This first migrates the code over to Amplify APIPlugin, and then any changes required over there have been ported back over here. Since all files under subscription folder is identical, we should be able to create a swift package of cocoapod out of those files.

## Major changes

- Some logic refactored into `BasicSubscriptionFactory`. BasicSubscriptionFactory is a class which creates connections/manages connection providers and is outside the Subscription folder (which will be turned into a package). This implementation on how to manage vend SubscriptionConnections is specific to AppSync's, such passing `authprovider` which create concrete classes (defined in AppSyncClient code) of interceptors for the message/connection request. For example, a request interceptor will have the user agent "appsync" rather than having an user agent like "amplify".

- `BasicSubscriptionFactory` now takes in the `authProvider` and creates an `authInterceptor`.
- It also holds onto a map from "auth type to auth interceptor", we can hold the map of "endpoint to connectionProvider" within the SubscriptionFactory. whenever we need a subscription connection for a particular URL and auth Type. 
The flow looks like:
1. retrieve the authInterceptor via the authTypeToInterceptor map.
2. retrieve or create the connection provider for the url (a check from the urlToConnectionProvider map)
3. Use the connection provider, and the auth interceptor, to create a new subscription connection on that connection provider aka websocket connection.


lawmicha:
https://github.com/aws-amplify/amplify-ios/pull/327

lawmicha:
See https://github.com/awslabs/aws-mobile-appsync-sdk-ios/pull/338 for latest

PR #338, opened 2020-02-19
Use AppSyncRealTimeClient CocoaPod

lawmicha:
This PR has a few commits which show the consolidation work so it can be reviewed.

The code was then moved over and checked in at https://github.com/aws-amplify/aws-appsync-realtime-client-ios

pod has been published as 1.0.0 to represent the current state of things without major bug fix or changes, except the call outs would be
- Interceptors are currently implemented by the consumer
- A subscription factory class is also implemenented by the consumer.
- both possible set of functionality that needs extra effort to be pulled into the shared package
- A lot of places required making classes public


AWSAppSync.podspec
```diff
@@ -14,7 +14,7 @@ Pod::Spec.new do |s|
   s.dependency 'AWSCore', '~> 2.12.0'
   s.dependency 'SQLite.swift', '~> 0.12.2'
   s.dependency 'ReachabilitySwift', '~> 5.0.0'
-  s.dependency 'Starscream', '~> 3.0.2'
```
royjit: Will this work without the Starscream?

AWSAppSyncClient/AuthInterceptor/AppSyncJSONHelper.swift
```diff
@@ -5,6 +5,7 @@
 //
 
 import Foundation
+import AppSyncSubscriptionClient
```
royjit: import here is needed?

Podfile
```diff
@@ -9,7 +9,8 @@ target "AWSAppSync" do
   pod "AWSCore", "~> #{AWS_SDK_VERSION}"
   pod "SQLite.swift", "~> 0.12.2"
   pod "ReachabilitySwift", "~> 5.0.0"
-  pod "Starscream", "~> 3.0.2"
+  pod "AppSyncSubscriptionClient", :path => '~/aws-amplify/amplify-ios'
```
royjit: Reminder to change this back

AWSAppSync.podspec
lawmicha: yup, since the dependency AppSyncRealTimeClient depends on Starscream. taking on the AppSyncRealTimeClient dependency is enough

AWSAppSyncClient/AuthInterceptor/AppSyncJSONHelper.swift
lawmicha: yeah, this is just a single method (base64AuthenticationBlog), that's used by two interceptors. `AuthenticationHeader` comes from AppSyncRealTimeClient.

Podfile
lawmicha: updated to use live pod!

jodafm:
Hi everyone.
I think you miss to update the cartfile adding "aws-appsync-realtime-client-ios" in order to add support to Carthage.
Please check my issue #363  for more information

Issue #342, opened 2020-02-25
Strong retain cycle of self inside of AppSyncSubscriptionWithSync


LachlanMcCulloch:
**Describe the bug**
When using any of the `AWSAppSyncClient` sync api, they return a `AppSyncSubscriptionWithSync` object, polymorphed as a `Cancellable` object. When calling `cancel()` and releasing the reference to this object I noticed that it would not deinitialize and persist to exist and continue its sync behaviour.

I discovered that within the `AppSyncSubscriptionWithSync` there are two points where `self` is taken as a strong reference. The first is in `performInitialSync()`:
```swift
...
internalStateSyncQueue.addOperation {
    self.registerForNotifications()
    self.initializeLastSyncTimeFromCache()
    self.initializeBaseQueryResultsFromCache()
    self.performSync()
}
...
```

and the second is in `performSync(at:)`:
```swift
...
nextSyncTimer = DispatchSource.makeOneOffDispatchSourceTimer(deadline: deadline, queue: handlerQueue) {
    AppSyncLog.debug("Timer fired, queueing sync operation")
    self.internalStateSyncQueue.addOperation {
        AppSyncLog.debug("Perform sync queued by timer")
        self.performSync()
    }
}
...
```

By updating both of these calls to take a weak self (`[weak self] in`), this removes the retain cycle and thus does not cause a memory leak when the returned `Cancellable` reference is released.

**To Reproduce**
Add the following unit test to existing master branch in `AWSAppSyncUnitTests/AppSyncSubscriptionWithSyncTests.swift`:
```swift
func testCancelWithBaseQueryAndDeinitNoStrongReferenceRetained() {
    var subscriptionWithSync: AppSyncSubscriptionWithSync? = AppSyncSubscriptionWithSync(
        appSyncClient: appsyncClient,
        baseQuery: listPostsQuery,
        deltaQuery: emptyQuery,
        subscription: emptySubscription,
        baseQueryHandler: listQueryHandler,
        deltaQueryHandler: emptyDeltaQueryHandler,
        subscriptionResultHandler: emptySubscriptionResultHandler,
        subscriptionMetadataCache: nil,
        syncConfiguration: .init(baseRefreshIntervalInSeconds: 1),
        handlerQueue: queue)

    weak var weakRef = subscriptionWithSync
    subscriptionWithSync?.cancel()
    subscriptionWithSync = nil
[...]

PR #343, opened 2020-02-25
Fix strong retain cycles

LachlanMcCulloch:
*Issue #342:*

*Description of changes:*
Takes a `weak self` reference inside of the initial sync operation and dispatch timer for sync.

The integration test  `testDeltaSyncMetadataClear` had to be updated to take a strong reference of the `Cancellable` objects.

The integration test `testStartStopStartSubscriptions` also had to be updated, although this appeared to be failing previous to the changes provided here.


LachlanMcCulloch:
Thanks for the speedy review @royjit ! Greatly appreciated!! :)

PR #350, opened 2020-02-28
Fixed the pod and updated the code to use the new code change

royjit:
- Fixed the internally connection to use optional data in variables
- Fixed podspec source files
- Updated the pod version in travis.yml


.travis.yml
```diff
@@ -12,6 +12,7 @@ before_install:
   -d
 - brew update
 - brew outdated carthage || brew upgrade carthage
+- gem install cocoapods -v '1.8.4'
```
lawmicha: is there a reason to hardcode a particular version here? when i was developing in this repo, i ended up updating cocoapods and bumped it to 1.9.0 when i did `pod install`

royjit: This is for travis CI. Travis was using "CocoaPods : 1.8.0" and there are few bug fixes in the latest cocoapod. So hardcoding it here will use the latest. 1.9.0+ is still in beta.

PR #366, opened 2020-03-13
Use AppSyncRealTimeClient 1.2.0 And Carthage fix

lawmicha:
This is pending AppSyncRealTimeClient 1.2.0 release (Pods and cartfile update, over in https://github.com/aws-amplify/aws-appsync-realtime-client-ios/pull/9)

After AppSyncRealTimeClient 1.2.0 is tagged/released then this PR needs to be updated with
1. `pod install` to get the 1.2.0 pod files under `Pod/` since they are checked in
2. Test out Carthage process to ensure consumers can use AppSync which depends on AppSyncRealTimeClient, and no longer StarScream.

### DIFF
--- a/Cartfile
+++ b/Cartfile
@@ -1,4 +1,4 @@
 github "aws/aws-sdk-ios" ~> 2.13.0
 github "stephencelis/SQLite.swift" ~> 0.12.2
 github "ashleymills/Reachability.swift" ~> 5.0.0
-github "daltoniam/starscream" ~> 3.0.2
+github "aws-amplify/aws-appsync-realtime-client" ~> 1.2.0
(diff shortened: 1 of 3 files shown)


Cartfile
```diff
@@ -1,4 +1,4 @@
 github "aws/aws-sdk-ios" ~> 2.13.0
 github "stephencelis/SQLite.swift" ~> 0.12.2
 github "ashleymills/Reachability.swift" ~> 5.0.0
-github "daltoniam/starscream" ~> 3.0.2
```
lawmicha: tried installing carthage dependencies in new project with just this line
issues because repo is not found, line needs to be updated to:

github "aws-amplify/aws-appsync-realtime-client-ios"

lawmicha:
closing for https://github.com/awslabs/aws-mobile-appsync-sdk-ios/pull/367

PR #371, opened 2020-03-17
Downgrade to 1.8.4 in Podfile.lock to match travisCI installation version

lawmicha:
Saw a build failure for the 3.1.0 release during cocoapods
https://travis-ci.org/github/awslabs/aws-mobile-appsync-sdk-ios/builds/663252569

```
NoMethodError - undefined method `url' for nil:NilClass
/Users/travis/.rvm/gems/ruby-2.6.4/gems/cocoapods-trunk-1.4.0/lib/pod/command/trunk/push.rb:155:in `master_repo_url'
```

Don't think the version mismatch is the issue, but could be, if the error message isn't very accurate.

travisCI installs cocoapods 1.8.4


lawmicha:
semi-related, https://github.com/awslabs/aws-mobile-appsync-sdk-ios/pull/350

PR #410, opened 2020-09-04
Upgrade to SDK 2.16.0

lawmicha:
*Issue #, if available:*

*Description of changes:*
Upgrade to SDK 2.16.0

*Testing done*
Running unit test and integration test on Xcode 12


palpatim (commented):
Looks good as far as I understand it, but do we know why the .xcconfig files for some pods were removed? Did you use the same version of CocoaPods to update the dependencies as we're going to be building with?

lawmicha:
> Looks good as far as I understand it, but do we know why the .xcconfig files for some pods were removed? Did you use the same version of CocoaPods to update the dependencies as we're going to be building with?

Ah this is because i did a re-install by removing existing `Pods/` and `Podfile.lock` then `pod install`. I'm going to revert this by just running `pod cache clean --all && pod install --repo-update` and it shouldn't delete those xcconfig files. I am using the same cocoapods version

Cartfile
```diff
@@ -1,4 +1,4 @@
-github "aws/aws-sdk-ios" ~> 2.14.0
+github "aws/aws-sdk-ios" ~> 2.16.0
```
palpatim, line 1: Eek, thanks for catching this.

We keep missing dependencies in the Cartfile, though. We need a mechanism to double-check that the Cartfile deps match those in the Podfile.

palpatim (approved):

Cartfile
lawmicha: we only ever bump this on minor version releases of the SDK, which is followed by an AppSync release so I got to this point from looking at the iOS release docs, and then over to the AppSync release docs. One place i can keep as a reminder is in the AppSync release docs, like "If this AppSync release is followed by the SDK minor release, do this" and have a reference to one of these PRs

PR #415, opened 2020-09-29
Added support for external URLSession handling for having things like certificate pinning

lesmuc:
*Description of changes:*
Added another init function that allows setting the URL Session from external. This will bring the possibility to implement the URLSessionDelegate outside the AWS Mobile AppSync SDK sources.

### DIFF
--- a/AWSAppSyncClient/AWSAppSyncHTTPNetworkTransport.swift
+++ b/AWSAppSyncClient/AWSAppSyncHTTPNetworkTransport.swift
@@ -25,6 +25,26 @@ public class AWSAppSyncHTTPNetworkTransport: AWSNetworkTransport {
     ///
     /// - Parameters:
     ///   - url: The URL of a GraphQL server to connect to.
+    ///   - apiKeyAuthProvider: An object of `AWSAPIKeyAuthProvider` protocol for API Key based authorization.
+    ///   - session: the URL session
+    ///   - sendOperationIdentifiers: Whether to send operation identifiers rather than full operation text, for use with servers that support query persistence. Defaults to false.
+    public init(url: URL,
+                apiKeyAuthProvider: AWSAPIKeyAuthProvider,
+                session: URLSession,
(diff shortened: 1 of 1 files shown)


palpatim:
I agree with the use case, but I think I'd like to do this as part of a larger refactor. This class is a bit of a mess, and adding your initializer to just the "API Key" authentication type leaves a gap, (not to mention that the proposed initializer doesn't include a `retryStrategy` argument :) ).

I'm working on a larger PR with a new initializer like so:

```swift
public init(
    url: URL,
    urlSession: URLSession,
    authProvider: AppSyncAuthProvider,
    sendOperationIdentifiers: Bool,
    retryStrategy: AWSAppSyncRetryStrategy
)
```

Where `AppSyncAuthProvider` is an internal enum of AWSAppSyncHTTPNetworkTransport, that matches closely with the existing `AWSAppSyncAuthType`:

```swift
public enum AppSyncAuthProvider {
    case awsIAM(provider: AWSCredentialsProvider, endpoint: AWSEndpoint)
    case apiKey(AWSAPIKeyAuthProvider)
    case oidcToken(AWSOIDCAuthProvider)
    case amazonCognitoUserPools(AWSCognitoUserPoolsAuthProvider)
}
```

So to solve your use case, you'd do something like:

```swift
let session = URLSession(...with my configuration and delegate and queue)
let transport = AWSAppSyncHTTPNetworkTransport(
        url: url,
        urlSession: session,
        authProvider: .apiKey(myAPIKeyProvider),
        sendOperationIdentifiers: true,
        retryStrategy: .exponential
    )
```

Thoughts?

lesmuc:
> 
> Thoughts?

Sorry for my late response and thanks for your response. Your proposal / PR looks good for me!

palpatim:
#422 released on https://github.com/awslabs/aws-mobile-appsync-sdk-ios/releases/tag/3.1.11

PR #418, opened 2020-10-09
Require AWS SDK 2.17.0

jdotw:
*Issue #, if available:*

*Description of changes:*

AWSMobileClient etc has updated to 2.17.0 which includes a fix required to build on Xcode 12.0.1. However, you can't install AWSAppSync 3.1.6 specified 2.16.0 creating a conflict where you can't build AWSAppSync on Xcode 12.0.1 because it isn't dependency-compatible with AWSCore 2.17.0. 

This change updates the Podfile, podspec and Cartfile to require AWSCore / SDK 2.17.0


palpatim:
As you've probably noticed, we've updated the dependencies separately as part of normal operations, so this is no longer necessary. Thanks for taking the time to submit the PR!

PR #426, opened 2020-10-16
fix: remove inconsistent test in AppSyncSubscriptionWithSync

lawmicha:
*Issue #, if available:*

*Description of changes:*
This test is inconsistent because `AppSyncSubscriptionWithSync` performs sync when it is initialized. Calling `cancel()` on it will be a race condition for whoever gets to the base query first or the check for cancelled first.

Related PR for cancel logic: https://github.com/awslabs/aws-mobile-appsync-sdk-ios/pull/355


lawmicha:
Instead of removing the test, find a way to test cancel(), such as asserting no more events occurring on the subscription events handler

PR #452, opened 2021-01-15
chore: fix testCancel() to make consistent

wooj2:
This unit test is inconsistent. So I think we have a couple options..

Option 1.  Just comment out the test, because there is clearly a race condition between an operation that is scheduled to start during instantiation and calling `cancel()`.  This is the approach implemented here:
https://github.com/awslabs/aws-mobile-appsync-sdk-ios/pull/426

Option 2. Implemented in this PR -- Allow the operation to start, cancel it, and then assert the state to make sure we have canceled the operation.  This seems like a reasonable approach, but introduces some risk since we are changing isCancelled to an atomic (that being said, it is the right thing to do since it can be accessed by multiple threads.

Overall the code isn't set up to be testable, so I think this option strikes a good balance between refactoring everything, and having some value in knowing we are properly calling cancel() on the operation queue.

### DIFF
--- a/AWSAppSyncClient/Internal/AppSyncSubscriptionWithSync.swift
+++ b/AWSAppSyncClient/Internal/AppSyncSubscriptionWithSync.swift
@@ -5,6 +5,7 @@
 //
 
 import Foundation
+import AppSyncRealTimeClient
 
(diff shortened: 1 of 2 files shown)


AWSAppSyncClient/Internal/AppSyncSubscriptionWithSync.swift
```diff
@@ -5,6 +5,7 @@
 //
 
 import Foundation
+import AppSyncRealTimeClient
```
lawmicha, line 8: Is this being used somewhere?

PR #453, opened 2021-01-20
Update Framework Application Extension API Build Settings

dlevans2:
* Update APPLICATION_EXTENSION_API_ONLY to YES to restrict API usage to those appropriate for an extension

*Issue #, if available:*

*Description of changes:*
Set APPLICATION_EXTENSION_API_ONLY build setting to YES for the framework target.  Including the framework in an extension, produces a `warning: linking against dylib not safe for use in application extensions` warning otherwise, but it doesn't look like the framework uses any API that would need the setting set to NO. 


atierian:
Thank you for opening this PR. 

AWS AppSync SDK for iOS entered maintenance mode in September 2023 and will receive no further updates as of September 2024.

PR #460, opened 2021-02-08
Make properties non-optional

muukii:
I found properties that no need to be optional type.


jcjimenez (approved):
LGTM

royjit:
Had to revert the commit since the changes are breaking for existing users.

PR #466, opened 2021-03-24
chore: Support Swift Package Manager

royjit:
*Issue #, if available:*

*Description of changes:*
- Removed objc version of AppsyncLogHelper and replaced with swift code
- Added unit test for the new AppsyncLogHelper
- Added missing imports
- Added Swift Package Manifest file

---

### Cocoapod

Currently cocoapod is used as the dependency management tool for development. We are not changing that as part of SPM support. Open the AWSAppSyncClient.xcworkspace and build the app should work (pod update if required). Also testing against a sample app and invoking Appsync apis should work.

- [x] Build Cocoapod workspace in AWSAppSync code
- [x] Test cases passed on running
- [x] Test against Sample app.

### SPM

SPM is added as the new distribution mechanism. We need to make sure that AWSAppSyncClient project builds correctly with the new SPM manifest file. Also testing against a sample app and invoking apis should work.

- [x] Build SPM manifest in AWSAppSync code
- [x] Test against Sample app.

Sample app integration steps:

1. Create an xcode project
1. Go to File → Swift Packages → Add Package dependency
1. Add the github url and select version

### Carthage

Carthage will be another distribution mechanism for AppSync. We need to make sure that using AWSAppSyncClient we can build and archive using Carthage. Also testing against a sample app and invoking apis should work.

- [x] Build Carthage in AWSAppsSync code ` ./build-support/carthage-build.sh build --no-skip-current --use-xcframeworks`
- [x] Test against Sample app.  ` carthage update --use-xcframeworks`
Cartfile used for sample app:

```

git "file:///<LOCAL PATH TO>/aws-mobile-appsync-sdk-ios" "<TEST BRANCH>"

```


AWSAppSyncClient/Internal/AppSyncLogHelper.swift
```diff
@@ -0,0 +1,34 @@
+//
```
palpatim, line 1: Thanks for simplifying this!

palpatim (commented):
Thanks for simplifying the log wrapper. How are we going to build the project as a swift-only SPM package while still maintaining Carthage compatibility (which requires us to commit the ObjC SDK sources)?

palpatim (approved):

palpatim (approved):

Package.swift
```diff
@@ -0,0 +1,32 @@
...
+
+let package = Package(
+    name: "AWSAppSync",
+    products: [
+        .library(
+            name: "AWSAppSync",
+            targets: ["AWSAppSync"]),
+    ],
+    dependencies: [
+        .package(name: "AWSiOSSDKV2", url: "https://github.com/aws-amplify/aws-sdk-ios-spm.git", from: "2.23.3"),
```
palpatim, line 14:
```suggestion
        .package(name: "AWSiOSSDKV2", url: "https://github.com/aws-amplify/aws-sdk-ios-spm.git", .upToNextMinor(from: "2.23.3")),
```

palpatim (changes requested):

palpatim (approved):

PR #488, opened 2021-09-23
fix: lambda async auth provider

diegocstn:
*Description of changes:*
Fix `LambdaAuthInterceptor` to call either the sync or async provider (AWSLambdaAuthProviderAsync) to retrieve the authorization token.

### DIFF
--- a/AWSAppSyncClient/Internal/AuthInterceptor/LambdaAuthInterceptor.swift
+++ b/AWSAppSyncClient/Internal/AuthInterceptor/LambdaAuthInterceptor.swift
@@ -74,6 +74,7 @@ class LambdaAuthInterceptor: AuthInterceptor {
         }
         var authToken: String?
         var authTokenError: Error?
+        let result: Swift.Result<String, Error>
         
         let semaphore = DispatchSemaphore(value: 0)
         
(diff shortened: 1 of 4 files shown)


lawmicha (approved):

AWSAppSyncClient/Internal/AuthInterceptor/LambdaAuthInterceptor.swift
```diff
@@ -61,6 +67,34 @@ class LambdaAuthInterceptor: AuthInterceptor {
...
+            semaphore.signal()
+        }
+        semaphore.wait()
+        
+        if let authTokenError = authTokenError {
+            result = .failure(authTokenError)
+        } else if let authToken = authToken {
+            result = .success(authToken)
+        } else {
+            fatalError("Incompatible values for authorization token and error: nil, nil")
```
lawmicha, line 93: is `assert` the right thing to do here? if the problem is just returning _something_, can return our own failure
```
assert(true, "Incompatible values for authorization token and error: nil, nil")
result = .failure(TokenError.unknown)
```

lawmicha (approved):

Issue #487, opened 2021-09-15
Crash in timer handling during retries in AWSAppSyncHTTPNetworkTransport (swift_isUniquelyReferenced_nonNull_native)


markusloffler:
**Describe the bug**
Crash in AWSAppSyncHTTPNetworkTransport, apparently this line:

```
AWSAppSyncHTTPNetworkTransport.sendGraphQLRequest(...)
{
     (...)
      self?.activeTimers.removeValue(forKey: taskUUID)
     (...)
}
```

This is the stack trace of the crash:

```
BACKGROUND THREAD 65 - CRASHED

libswiftCore.dylib:
swift_isUniquelyReferenced_nonNull_native

Connected:
generic specialization <Swift.String, __C.OS_dispatch_source_timer>
of Swift.Dictionary._Variant.removeValue(forKey: A) -> B?

Connected:
closure #1 () -> ()
in closure #1 (Swift.Result<[Swift.String : Any], AWSAppSync.AWSAppSyncClientError>) -> ()
in closure #1 (Swift.Result<(), Swift.Error>) -> ()
in AWSAppSync.AWSAppSyncHTTPNetworkTransport.sendGraphQLRequest(mutableRequest: __C.NSMutableURLRequest,
     retryHandler: AWSAppSync.AWSAppSyncRetryHandler, networkTransportOperation:
     AWSAppSync.AWSAppSyncHTTPNetworkTransport.AWSAppSyncHTTPNetworkTransportOperation,
     completionHandler: ([Swift.String : Any]?, AWSAppSync.AWSAppSyncClientError?) -> ()) -> ()

Connected:
reabstraction thunk helper from @escaping @callee_guaranteed () -> ()
to @escaping @callee_unowned @convention(block) () -> ()

libdispatch.dylib:       _dispatch_client_callout
libdispatch.dylib:       _dispatch_continuation_pop
libdispatch.dylib:       _dispatch_source_invoke
libdispatch.dylib:       _dispatch_root_queue_drain
libdispatch.dylib:       _dispatch_worker_thread2
libsystem_pthread.dylib: _pthread_wqthread
libsystem_pthread.dylib: start_wqthread
```

**To Reproduce**
The crash happened during a time when the backend was down and throwing 500 errors, so AWSAppSyncHTTPNetworkTransport started retries. Unforuntately, I could never reproduce it at will, it's something sporadic.

We noticed the crash only through our crash reporter.

**Expected behavior**
No crash

[...]

PR #501, opened 2021-11-12
fix: remove race condition on activeTimer access in AWSAppSyncHTTPNetworkTransport

lawmicha:
*Issue #, if available:*
https://github.com/awslabs/aws-mobile-appsync-sdk-ios/issues/487

*Description of changes:*
`activeTimers` is of type `[String: DispatchSourceTimer]` so accessing this does not appear to be thread safe in the code. `sendGraphQLRequest` under certain conditions will retry like when service returns 5xx (or other retry scenarios). It will create a new timer, set it on `activeTimers`, while other timers that were created already may fire and remove it's timer from `activeTimers`. A race condition could happen if T1 calls sendGraphQLRequest due to a previous timer firing, fails the request, completes on T2 and retry logic decides to create another timer which accesses `activeTimers`. T1 continues to the next line from `sendGraphQLRequest` and removes a timer, which happens at the same time T2 is setting the new timer.

This PR adds the access to `activeTimers` under a [Dispatch Barrier](https://developer.apple.com/documentation/dispatch/dispatch_barrier) to ensure that each block of execution done serially
- [x] Ran unit and integration tests

### DIFF
--- a/AWSAppSyncClient/AWSAppSyncHTTPNetworkTransport.swift
+++ b/AWSAppSyncClient/AWSAppSyncHTTPNetworkTransport.swift
@@ -6,6 +6,7 @@
 
 import Foundation
 import AWSCore
+import AppSyncRealTimeClient
 
 public class AWSAppSyncHTTPNetworkTransport: AWSNetworkTransport {
     public enum AppSyncAuthProvider {
@@ -37,7 +38,8 @@ public class AWSAppSyncHTTPNetworkTransport: AWSNetworkTransport {
     private let authProvider: AppSyncAuthProvider
     private let sendOperationIdentifiers: Bool
     private var retryStrategy: AWSAppSyncRetryStrategy
-
+    
+    private let activeTimersQueue = DispatchQueue(label: "AWSAppSyncHTTPNetworkTransport.activeTimers", attributes: .concurrent)
     private var activeTimers: [String: DispatchSourceTimer] = [:]
 
     /// Designated initializer. Creates a network transport with the specified server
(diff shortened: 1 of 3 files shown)


AWSAppSyncClient/AWSAppSyncHTTPNetworkTransport.swift
```diff
@@ -6,6 +6,7 @@
 
 import Foundation
 import AWSCore
+import AppSyncRealTimeClient
```
lawmicha, line 9: will be removing this `import` as it is unsed, was trying out the `AtomicValue` type that we have in AppSyncRealTimeClient that uses the NSLock earlier

AWSAppSyncClient/AWSAppSyncHTTPNetworkTransport.swift
```diff
@@ -249,9 +250,13 @@ public class AWSAppSyncHTTPNetworkTransport: AWSNetworkTransport {
                                                          retryHandler: retryHandler,
                                                          networkTransportOperation: networkTransportOperation,
                                                          completionHandler: completionHandler)
-                                self?.activeTimers.removeValue(forKey: taskUUID)
+                                self?.activeTimersQueue.async(flags: .barrier) {
```
brennanMKE, line 253: What I find works best is to have private functions which handle protected state transitions which you can sync. With private functions you know there is no way to call those functions from outside this type, so you can strictly control the context of the work. I like to use `dispatchPrecondition` to ensure a private function is not on the internal queue before calling `queue.sync` in it. And I want to use `sync` instead of `async` if it is just a small task like changing the value of a stored property. For a longer running task I would want to make sure it is run off the `.main` queue by moving the work to `global()` and then syncing on the internal queue. You never want to block the main thread for too long as it risk the watchdog killing the app.

With private functions you can also ensure these functions to do not call each other within the sync block. And if you only use the queue in these private functions and have `dispatchPrecondition` guarding the context when entering the functions you will make sure you cannot reach a deadlock. 

When you use `sync` you also know the stored property is set by the time execution reaches the next line.

brennanMKE: I've prepared this [PR](https://github.com/awslabs/aws-mobile-appsync-sdk-ios/pull/503) with these changes on top of your changes. The unit tests are passing.

brennanMKE (commented):
Added comments

AWSAppSyncClient/AWSAppSyncHTTPNetworkTransport.swift
lawmicha: thanks, merged https://github.com/awslabs/aws-mobile-appsync-sdk-ios/pull/503 into this branch

brennanMKE (commented):
LGTM

PR #503, opened 2021-11-30
Stehlib sync functions

brennanMKE:
*Issue #, if available:*

#501 

This is an update on top of another [PR](https://github.com/awslabs/aws-mobile-appsync-sdk-ios/pull/501).

### DIFF
--- a/AWSAppSyncClient/AWSAppSyncHTTPNetworkTransport.swift
+++ b/AWSAppSyncClient/AWSAppSyncHTTPNetworkTransport.swift
@@ -6,6 +6,7 @@
 
 import Foundation
 import AWSCore
+import AppSyncRealTimeClient
 
 public class AWSAppSyncHTTPNetworkTransport: AWSNetworkTransport {
     public enum AppSyncAuthProvider {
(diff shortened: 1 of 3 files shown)


AWSAppSyncClient/AWSAppSyncHTTPNetworkTransport.swift
```diff
@@ -511,6 +518,28 @@ public class AWSAppSyncHTTPNetworkTransport: AWSNetworkTransport {
...
 
+    // MARK: - Protected State Transitions -
+
+    private func removeActiveTimer(taskUUID: String) {
+        if #available(iOS 10.0, *) {
+            dispatchPrecondition(condition: .notOnQueue(activeTimersQueue))
+        }
+        activeTimersQueue.sync {
+            _ = activeTimers.removeValue(forKey: taskUUID)
+        }
```
lawmicha, line 529: is there a reason why we are no longer using `.async(flags: .barrier)` and using `sync` block? The reason why I wanted async barrier is because the caling code can send off the work using `async` block and guaranteed that the work performed in the block is performed one after another since accessing `activeTimers` is not thread safe. If my understanding is correct here, the change to `sync` block, means that the calling code will block on this method and continue only after it's done. 

The calling code, which is to perform a network request, fails, and should retry is true, will then create a timer, and store the timer, and return. Seems like storing the timer under `async barrier` or `sync` will achieve the same thing. Both actions are non-blocking from the app's perspective. The question is, do we need to wait until the timer is stored (using sync block) or can the calling code finish immediately (using async barrier)?

lawmicha: i see you've explained reason for using `sync` over `async` here https://github.com/awslabs/aws-mobile-appsync-sdk-ios/pull/501#discussion_r759593510

lawmicha (approved):
Thanks for reviewing by preparing this PR! changes LGTM. If you like we can merge this PR into main and close off the previous one (https://github.com/awslabs/aws-mobile-appsync-sdk-ios/pull/501) . Updating PR title

PR #523, opened 2022-02-25
fix: Thread safe subscription watcher cancel

lawmicha:
*Issue #, if available:*

*Description of changes:*


AWSAppSyncClient/AWSAppSyncSubscriptionWatcher.swift
```diff
@@ -177,22 +179,26 @@ public final class AWSAppSyncSubscriptionWatcher<Subscription: GraphQLSubscripti
...
     /// `resultHandler`, although it will set the internal state of the watcher to `.disconnected`
     public func cancel() {
-        if self.cancellationSource == .none {
-            self.cancellationSource = .user
+        if !self.isCancelled {
+            if self.cancellationSource == .none {
+                self.cancellationSource = .user
+            }
+            self.performCleanUpTasksOnCancel()
         }
```
diegocstn, line 187: nit: can we use a `guard` and an early return instead of nested ifs?

diegocstn (approved):

Issue #92, opened 2018-10-31
Optimistic update fails if cache is empty


freshtechs:
**State your question**
Optimistic update always fail to "missingValue", if model wasn't previously cached from a fetch query. And said fetch operation needs to actually populate the cache with items, for further optimistic updates, otherwise (i.e. cache is empty) the optimistic update won't update/initialize the cache(?)

Don't know how could I initialize an empty cache just in case the **first** mutation fails to reach backend, and needs to be written on cache/localDB

**Environment(please complete the following information):**
 - AppSync SDK Version: 2.6.22
 - Dependency Manager:  CocoaPods
 - Swift Version : 4.2

**Device Information (please complete the following information):**
 - Device: Simulator
 - iOS Version: iOS 12

sdeff:
I'm facing the same problem.

When I run the Event App sample for the very first time and with airplane mode enabled, I get a "missingValue" after executing "transaction?.update(...)" whenever I try to add a new post:

        appSyncClient?.perform(mutation: addEventMutation, optimisticUpdate: { (transaction) in
            do {
                // Update our normalized local store immediately for a responsive UI.
                try transaction?.update(query: ListEventsQuery()) { (data: inout ListEventsQuery.Data) in
                    data.listEvents?.items?.append(
                        ListEventsQuery.Data.ListEvent.Item(id: UUID().uuidString,
                                                            description: descriptionText,
                                                            name: nameText,
                                                            when: whenText,
                                                            where: whereText,
                                                            comments: nil))
                }
            } catch {
                print("Error updating the cache with optimistic response.")
            }
        }

However, when I force close the app and start it with airplane mode disabled, the previously added posts appear.

freshtechs:
This may be a bug? @joelmbell

freshtechs:
This may be a bug? @rohandubal

rohandubal:
@sdeff @freshtechs There was a known bug which was fixed in 2.6.22 which did not allow writing to cache for queries with parameters or with objects with null values.

Does this happen only on queries which have never been fetched from the network or for any query?

freshtechs:
@rohandubal Happens when you try to perform an optimistic update mutation prior a fetch request of the same object/model (from cache or network i tried with all the keys in CachePolicy).

 If the first action and first launch of your app involves an optimistic update the perform operation would fail to the “do try transaction?.update...” catch statement printing “missingValue” as error value

rohandubal:
[...]

PR #535, opened 2022-05-16
fix: Repopulate records table with empty QUERY_ROOT after caches are cleared

jellybeansoup:
*Issue #, if available:*
#92 (kinda?)

*Description of changes:*
As part of working around an issue allowing optimistic updates of query results that haven't been retrieved from the service, @palpatim introduced code to pre-populate an empty table with an empty QUERY_ROOT, resolving the issue. However the issue resurfaces if at any point the caches are cleared, as the table is not then repopulated with the empty QUERY_ROOT.

So if, for example, a user installs an app for the first time and log into an account backed by AWS App Sync, the existing workaround correctly ensures that optimistic updates don't fail when the underlying query results have not been fetched from the service, but if that user were to sign out (causing the caches to be cleared) and then into another account, the workaround is no longer in place, and optimistic updates will fail.

This PR introduces changes to repopulate the cache with the empty QUERY_ROOT immediately after clearing it, thereby ensuring that optimistic updates do not fail after calling `clearCaches()`. The requisite tests have also been updated to showcase the issue, and show that it is resolved.

### DIFF
--- a/AWSAppSyncClient/AWSAppSyncClientConfiguration.swift
+++ b/AWSAppSyncClient/AWSAppSyncClientConfiguration.swift
@@ -587,13 +587,7 @@ public class AWSAppSyncClientConfiguration {
         if let databaseURL = databaseURL, let cache = try? AWSSQLiteNormalizedCache(fileURL: databaseURL) {
             store = ApolloStore(cache: cache)
         } else {
-            // Prepopulate the InMemoryNormalizedCache record set with an empty QUERY_ROOT, to allow optimistic
-            // updates against empty caches to succeed. Otherwise, such an operation will fail with a "missingValue"
-            // error (#92)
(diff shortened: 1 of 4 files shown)


atierian:
Thanks for opening this PR @jellybeansoup. We'll review and get back to you on this shortly.

lawmicha (changes requested):
Thanks @jellybeansoup for opening this PR, I ran the integration tests and they are passing. Suggestion to refactor some of the code here: https://github.com/awslabs/aws-mobile-appsync-sdk-ios/commit/1552e351c2ed38d545d3bab928a910f07027b36b please cherry-pick or apply a similar change before we merge this in

jellybeansoup:
Thanks so much, @lawmicha. I've cherry-picked the requested changes; really should've picked up on that myself 😅

lawmicha (approved):

PR #536, opened 2022-05-17
fix: deserialization of custom scalars

jellybeansoup:
*Issue #, if available:*
None.

*Description of changes:*
We use a number of custom scalars with AWS AppSync in our codebase, which are successfully parsed from the AppSync service, and get stored within the SQLite cache, but fail to be deserialised and thus cause cache reads involving these types to completely fail. This has primarily meant that we are unable to cache these calls at all, relying on live data for calls relying on our custom scalar types.

This is exactly the same as [an issue raised on the Apollo library in 2018](https://github.com/apollographql/apollo-ios/issues/329), and was resolved by returning the `fieldJSONValue` from the `SQLiteSerialization.deserialize(fieldJSONValue:)`, which correctly returns the custom scalar value as-is ([in this PR](https://github.com/apollographql/apollo-ios/pull/1144)).

I've replicated the changes here, including the removal of the then defunct `AWSAppSyncQueriesCacheError.invalidRecordValue` case. 

### DIFF
--- a/AWSAppSyncClient/AWSAppSyncCache.swift
+++ b/AWSAppSyncClient/AWSAppSyncCache.swift
@@ -10,6 +10,8 @@ import Foundation
 public enum AWSAppSyncQueriesCacheError: Error {
     case invalidRecordEncoding(record: String)
     case invalidRecordShape(object: Any)
+    @available(*, deprecated, message: "No longer being used and will be removed in a future revision.")
+    case invalidRecordValue(value: Any)
 }
 
 /// Errors thrown during creation or migration of AppSync caches
--- a/AWSAppSyncClient/AWSAppSyncCache.swift
+++ b/AWSAppSyncClient/AWSAppSyncCache.swift
@@ -10,7 +10,6 @@ import Foundation
 public enum AWSAppSyncQueriesCacheError: Error {
     case invalidRecordEncoding(record: String)
     case invalidRecordShape(object: Any)
-    case invalidRecordValue(value: Any)
 }
 
 /// Errors thrown during creation or migration of AppSync caches
--- a/AWSAppSyncClient/Internal/SQLiteSerialization.swift
+++ b/AWSAppSyncClient/Internal/SQLiteSerialization.swift
@@ -45,7 +45,7 @@ final class SQLiteSerialization {
         switch fieldJSONValue {
         case let dictionary as JSONObject:
             guard let reference = dictionary[serializedReferenceKey] as? String else {
-                throw AWSAppSyncQueriesCacheError.invalidRecordValue(value: fieldJSONValue)
+                return fieldJSONValue
             }
             return Reference(key: reference)
         case let array as [JSONValue]:


atierian:
Thanks for opening this PR (and the other one). We'll review and come back to you on this.

AWSAppSyncClient/Internal/SQLiteSerialization.swift
```diff
@@ -45,7 +45,7 @@ final class SQLiteSerialization {
         switch fieldJSONValue {
         case let dictionary as JSONObject:
             guard let reference = dictionary[serializedReferenceKey] as? String else {
-                throw AWSAppSyncQueriesCacheError.invalidRecordValue(value: fieldJSONValue)
+                return fieldJSONValue
```
lawmicha, line 48: From the Apollo issue linked in the description, has the PR with the same change to the `SQLiteSerialization` file here https://github.com/apollographql/apollo-ios/pull/1144/files#diff-e9bbcdac7903360aed17a5df82e644121e800adbc8b56bc03ffcf1dccc7dfe0a . With this, this change looks accurate to do since I think the file was lifted from the apollo codebase.

AWSAppSyncClient/AWSAppSyncCache.swift
```diff
@@ -10,7 +10,6 @@ import Foundation
 public enum AWSAppSyncQueriesCacheError: Error {
     case invalidRecordEncoding(record: String)
     case invalidRecordShape(object: Any)
-    case invalidRecordValue(value: Any)
```
lawmicha, line 13: can this be added back in? changes to the public interfaces will break existing customers or we will have to do a major version bump. Unless absolutely necessary, I think we should keep the enum case even if it's no longer used and in the future remove it completely in a major version bump. The incremental release will let existing developers just upgrade to a minor version to take the new behavior but not have to change the code that uses this enum

lawmicha (commented):
Thanks for opening this PR @jellybeansoup. Let's verify integration tests are passing as expected (SDK owners can assist in running this if it's difficult for developer contributor to perform this)

AWSAppSyncClient/AWSAppSyncCache.swift
jellybeansoup: I can absolutely put it back, and totally understand the reasoning. Would you like me to mark that case as `@available(*, deprecated)`, since it's no longer used or thrown from anywhere?

jellybeansoup:
> Let's verify integration tests are passing as expected (SDK owners can assist in running this if it's difficult for developer contributor to perform this)

I'd certainly appreciate it if someone could assist; I had trouble getting the appropriate AWS environment set up (the lambda instructions in particular), but I did verify that the other tests run and pass as expected.

AWSAppSyncClient/AWSAppSyncCache.swift
lawmicha: sure that makes sense, thanks! Was thinking something like this:
```
@available(*, deprecated, message: "No longer being used and will be removed in a future revision.")
```

lawmicha:
> > Let's verify integration tests are passing as expected (SDK owners can assist in running this if it's difficult for developer contributor to perform this)
> 
> I'd certainly appreciate it if someone could assist; I had trouble getting the appropriate AWS environment set up (the lambda instructions in particular), but I did verify that the other tests run and pass as expected.

Ran the integration tests and they're all passing

AWSAppSyncClient/AWSAppSyncCache.swift
jellybeansoup: Done and done! 😊

jellybeansoup:
Thanks for running the integration tests on my behalf, @lawmicha! I've cherry-picked the commit to return the removed error case as suggested, so it should be good for another look 😊

lawmicha (approved):

PR #542, opened 2022-06-27
Fix: release the session

bhushanever:
*Issue #, if available:*

*Description of changes:*


bhushanever:
Hi @dpilch,

Actually, in my application, I am using **appsync-sdk-ios** so while calling API I am facing a memory leak issue, so we tested many times and we come up with this. 

Why do we change this:-
Since a new session is used for each request, we can safely allow the session to be invalidated after the request has been completed.

@dpilch Please approve this PR. 

Thanks in advance.

dpilch:
The change appears to cause the unit and integration tests to fail. Could you take a look and update the tests if necessary?

PR #545, opened 2022-07-14
chore: Migrate to CircleCI

lawmicha:
*Issue #, if available:*

*Description of changes:*
This PR takes https://github.com/awslabs/aws-mobile-appsync-sdk-ios/pull/469 and attempts to check if there's anything else missing to enable CircleCI 

Things remaining:
- carthage build (either document manual steps or add to circleCI)?
- pod push will fail without the environment variable. This is expected, so need to set up pod session add-token script and rotation
- integration tests are not running in the workflow, i think they were running before in TravisCI, or not. make sure release process documents that the integration tests should be run beforehand


build-support/create_carthage_archive.py
```diff
@@ -0,0 +1,14 @@
...
+
+from functions import log
+
+project_dir = os.getcwd()
+log(f"Creating Carthage archive in {project_dir}")
+
+xcframework_path = f"{project_dir}/xcframeworks/output/XCF/"
+log(f"XCFramework from -  {xcframework_path}")
+
+framework_file_name = "aws-sdk-ios-carthage.framework"
```
lawmicha, line 12: this looks inaccurate

PR #548, opened 2022-08-15
fix: Perform optimistic updates without awaiting

jellybeansoup:
*Issue #, if available:*
Within the SEEK application, there are occasions where performing a mutation's optimistic update would cause the main thread to lock up for long enough that our crash analytics considers it a hang (typically a couple of seconds), thereby crashing the application.

It's hard to determine an exact cause for this, and it's not at all helped by how rare the issue is. My educated guess is that something is attempting to come back to the main thread as part of performing the optimistic update, potentially within the SQLite layer that I don't have a lot of visibility into. Alternatively, it's possible that there is something like a file system delay that causes SQLite to wait for the file to be unlocked. Regardless, this means that performing a mutation on the main thread (fairly common in scenarios initiated by user interaction) mean that the main thread is locked, as `Promise.await()` is called synchronously, and thus a semaphore is created.

The resulting stacktrace (truncated to the relevant parts) looks something like this:
1. [Promise.swift, line 106](https://github.com/awslabs/aws-mobile-appsync-sdk-ios/blob/35d9c6ece50ba01fd42aaeffc67f34c19cec624e/AWSAppSyncClient/Apollo/Sources/Apollo/Promise.swift#L106)
1. [AWSAppSyncClient.swift, line 293](https://github.com/awslabs/aws-mobile-appsync-sdk-ios/blob/35d9c6ece50ba01fd42aaeffc67f34c19cec624e/AWSAppSyncClient/AWSAppSyncClient.swift#L293)

*Description of changes:*
In attempting to preserve the public ABI, I've opted to move the optimistic update into the mutation queue, and have removed a reliance upon the `await()` method. This allows the mutation operation to be created (so it may be returned as `Cancellable`), while ensuring it isn't actually added to the queue until after the optimistic update is performed.

This allows the method to perform a mutation to return as quickly as possible, and avoids blocking the main thread.

**Please note:** I have not been able to successfully run integration tests (similar to past contributions), however we have validated that the issue is resolved by shipping the SEEK app using a fork with this change in place, and can confirm that the issue is eradicated.

### DIFF
--- a/AWSAppSyncClient/AWSAppSyncClient.swift
+++ b/AWSAppSyncClient/AWSAppSyncClient.swift
@@ -111,6 +111,7 @@ public class AWSAppSyncClient {
         self.mutationQueue = AWSPerformMutationQueue(
             appSyncClient: self,
             networkClient: httpTransport!,
+            apolloStore: store!,
             reachabiltyChangeNotifier: NetworkReachabilityNotifier.shared,
             cacheFileURL: appSyncConfig.cacheConfiguration?.offlineMutations)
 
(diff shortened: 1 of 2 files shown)


royjit (approved):
LGTM

5d (approved):

Issue #549, opened 2022-09-16
SPM dependency conflict with Amplify library


Devepre:
**Describe the bug**
Xcode fails to add  **latest 3.6.1** version of [aws-mobile-appsync-sdk-ios](https://github.com/awslabs/aws-mobile-appsync-sdk-ios) library while **latest 1.28.1** version of [amplify-ios](https://github.com/aws-amplify/amplify-ios) library already added to the project.

**To Reproduce**
1. Create new project using Xcode
2. Using SPM add latest **1.28.1** version of [amplify-ios](https://github.com/aws-amplify/amplify-ios) library (exact version)
3. Using SPM add latest **3.6.1** version of [aws-mobile-appsync-sdk-ios](https://github.com/awslabs/aws-mobile-appsync-sdk-ios) library (exact version)
4. See error "unable to resolve dependencies"

**Expected behavior**
Library should be added to project without error.

**Additional context**
This happens because latest [aws-mobile-appsync-sdk-ios](https://github.com/awslabs/aws-mobile-appsync-sdk-ios) requires:
```
        .package(
            name: "AWSiOSSDKV2",
            url: "https://github.com/aws-amplify/aws-sdk-ios-spm.git",
            .upToNextMinor(from: "2.27.0")
        )
```
while latest [amplify-ios](https://github.com/aws-amplify/amplify-ios) requires:
```
        .package(
            name: "AWSiOSSDKV2",
            url: "https://github.com/aws-amplify/aws-sdk-ios-spm.git",
            .upToNextMinor(from: "2.28.0")
        )
```

Probably the easiest way to fix is to change requirement to be the same as _.upToNextMinor(from: "2.28.0")_ for the library.

fzy-spyro:
I am also having this issue and it block me from using latest amplify sdk where following issue was resolved: https://github.com/aws-amplify/aws-sdk-ios/issues/4153

fzy-spyro:
Hey, 
Any update on the above?

albemala:
PR by @alexanderedge: https://github.com/awslabs/aws-mobile-appsync-sdk-ios/pull/552

gevartosky:
I'm having the same issue but with amplify 2.0.0. We want to adopt the new Async/Await that amplify provides but it uses `aws-appsync-realtime-client-ios.git 2.1.1` while AppSync uses 1.8.0 😞

Ricardo1980:
PR to do the same in the podspec.
[...]

PR #553, opened 2022-12-18
Update AWSAppSync.podspec

Ricardo1980:
*Issue #, if available:*
https://github.com/awslabs/aws-mobile-appsync-sdk-ios/issues/549

*Description of changes:*
That issue and the attached PR is about SPM.
This PR does exactly the same but for Cocoapods.

So, we can now update other dependecies.


royjit:
No longer required

PR #558, opened 2023-01-24
fix: adapt to appsync client v3

5d:
*Issue #, if available:*

*Description of changes:*
- adapting dependency appsync client v3 APIs


lawmicha (changes requested):
Thanks for trying this out, looks like it's going to be pretty minimal changes. Let's run the integration tests as well, they should be passing but might not be compiling at first as there might be some test code updates needed.

From the prespective of the developer using AppSync SDK, I think even though the dependency is using AppSyncRT v3, this update can still be a patch update considering the V3 behavior hasn't changed. we can rename "feat" to "fix: adapt to appsync client v3" or "chore: adapt to appsync client v3", since "feat" should indicate a minor or major version bump

Issue #568, opened 2023-04-13
AWS subscriptions - subscribe function is not thread safe and causes crashes


Vanyaslav:
**Describe the bug**
When subscribing on multiple threads the connection pool is causing crashes since it accesses a dictionary inside the class IAMBasedConnectionPool which is not thread safe.

**To Reproduce**
1. Create multiple subscribers with AWS App sync on multiple threads.
2. Enjoy the crashing app.

**Expected behavior**
It should be a thread safe operation to avoid crashes.

**Screenshots**
<img width="930" alt="Screenshot 2023-04-13 at 09 52 07" src="https://user-images.githubusercontent.com/1846398/231700087-fba13762-558b-406c-b45e-4edb89949245.png">
<img width="930" alt="Screenshot 2023-04-13 at 10 02 37" src="https://user-images.githubusercontent.com/1846398/231700114-c2a7085f-1d8d-4215-bbe3-5b3acced6678.png">

**Environment(please complete the following information):**
 - AppSync SDK Version: [3.6.2]
 - Dependency Manager: [Cocoapods]
 - Swift Version : [5.0]

**Device Information (please complete the following information):**
 - Device: [Simulator]
 - iOS Version: [e.g. iOS 16.4]
 - Specific to simulators:

**Additional context**
We use the framework kinda extensively, but with iOS16.4 and AppSync v. 3.6.2 the problem has gone to really a bad situation, while before we had experienced the crash less frequently.
Already prepared a fix , gonna send a PR soon.

atierian:
Resolved by https://github.com/awslabs/aws-mobile-appsync-sdk-ios/pull/569

PR #569, opened 2023-04-24
Fix: serialized requests on IAMBasedConnectionPool

chris-trick:
dictionary is not a thread safe object, need to be serialized in some way to avoid crashes

*Issue #568*

*Description of changes:*


ruisebas (approved):

petracvrljevic:
Could this same be applied to APIKeyBasedConnectionPool?

